PR #3, opened 2020-06-23
Flight object

emaohn:
- Flight component displays all flight data in travel object
- Fixed some grid/spacing issue in TravelObject

### DIFF
--- a/step-capstone/src/components/Flight.js
+++ b/step-capstone/src/components/Flight.js
@@ -3,6 +3,6 @@ import { Box, Grid, Typography } from '@material-ui/core'
 
 export default function Flight(props) {
     return(
-        <Typography variant="h4" gutterBottom>Flight!</Typography>
+        <Typography variant="h5" gutterBottom>{ props.data.departureAirport } to { props.data.arrivalAirport }</Typography>
     )
 }
--- a/step-capstone/src/components/Flight.js
+++ b/step-capstone/src/components/Flight.js
@@ -1,8 +1,13 @@
 import React from 'react'
 import { Box, Grid, Typography } from '@material-ui/core'
 
-export default function Flight(props) {
+export default function Flight({ data }) {
     return(
-        <Typography variant="h5" gutterBottom>{ props.data.departureAirport } to { props.data.arrivalAirport }</Typography>
+        <Grid item>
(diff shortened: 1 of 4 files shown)


step-capstone/src/components/Flight.js
```diff
@@ -0,0 +1,26 @@
...
+          finalized: true,
+          type: "flight",
+          departureAirport: "BOS",
+          arrivalAirport: "SFO",
+          departureDate: "4:00pm EST",
+          arrivalDate: "7:00pm PST",
+          description: "Additional notes"
+        }
+
+*/
```
zirving, line 16: good idea to give a comment detailing the data format

step-capstone/src/components/Flight.js
```diff
@@ -0,0 +1,26 @@
...
+import { Box, Grid, Typography } from '@material-ui/core'
+
+/*
+Takes data in this form:
+        {
+          finalized: true,
+          type: "flight",
+          departureAirport: "BOS",
+          arrivalAirport: "SFO",
+          departureDate: "4:00pm EST",
```
zirving, line 11: this would also include the date (may 17th for example) correct?

emaohn: Hmm I was thinking since travel objects in the itinerary would only show up under that specific day we wouldn't need to display the date.

step-capstone/src/components/Flight.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { Box, Grid, Typography } from '@material-ui/core'
+
+/*
+Takes data in this form:
```
ganewman42, line 5: Where does this json come from? If it's a source where there's possible errors, consider adding a JSON schema (https://www.npmjs.com/package/jsonschema) so that you can validate the input. The schema then serves the same documentation purpose as your comment does now.

step-capstone/src/components/Flight.js
ganewman42: Remember that flights might depart and arrive on different dates, so you would need a way to note that if you don't store the departure and arrival date.

step-capstone/src/components/Flight.js
emaohn: That makes sense! The data should be passed down directly from the database in a higher level component.

PR #4, opened 2020-06-23
Hotel

zirving:
- remove unneccesary import of serve.js (i got this from some tutorial but it works without it)

- created Hotel component
image


step-capstone/src/Components/TravelObject.js
```diff
@@ -1,18 +1,19 @@
...
+    switch (props.data.type) {
         case 'event':
             content = <Typography variant="h4" gutterBottom>Event!</Typography>
             break;
         case 'flight':
             content = <Typography variant="h4" gutterBottom>Flight!</Typography>
             break;
         case 'hotel':
-            content = <Typography variant="h4" gutterBottom>Hotel!</Typography>
+            content = <Hotel {...props.data}/>
```
emaohn, line 16: What's the purpose of "..."? Would it maybe be clearer to pass it in like: data={ props.data }?

step-capstone/src/Components/Hotel.js
```diff
@@ -0,0 +1,26 @@
...
+   * checkOut:  December 20, 2020 11:00:00
+   * description: 'description'
+   * finalized : true
+  **/
+  return (
+    <Grid item>
+      <Typography variant="h6" gutterBottom>{data.name} </Typography>
+      <Typography variant="subtitle2" gutterBottom>{
+          data.checkIn + ' - ' +
+          data.checkOut} </Typography>
```
emaohn, line 21: Nit: Maybe move the closing tag </Typography> to the next line for style?

step-capstone/src/Components/TravelObject.js
zirving: ...props.data would deconstruct props.data, so that going into Hotel the data fields are listed seperately. so in Hotel you can just do props.title. If you do data = props.data you are creating a specific prop within props that contains all the data fields. so you would need to do props.data.title. It depends on whether we think anything else other than those data fields is getting passed to hotel.

step-capstone/src/Components/Hotel.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import {Grid, Typography} from '@material-ui/core'
+
+export default function Hotel(data) {
+
+  /**
+   * Takes data in the form: 
+   * type : 'hotel
+   * name : ' Hotel ZED ' 
+   * location : 'London'
```
ganewman42, line 10: For mapping purposes, it might be smart to store latitude and longitude as a unique identifier of a location. Hotels don't always have unique names or may not be registered to google maps, so looking up "Hotel ZED in London" might return zero or many results.

step-capstone/src/Components/Hotel.js
```diff
@@ -0,0 +1,26 @@
...
+   * checkIn : December 17, 2020 15:30:00
```
ganewman42, line 11: Consider including timezone here (if you're planning on including it in the spec go ahead and put it in the comment as well).

PR #6, opened 2020-06-24
Trip component

emaohn:
- layout for all child components (sizing, position, etc)
- Add button with popover to add information about a new item
- popover has tab to select hotel, event, flight and renders specific forms based on selection
- Brings new item data back up to trip component and adds to list when create button is pressed
- Splits list into two upon rendering, passing finalized items to the itinerary and unfinalized on the onorganized component.

- Blue space = map component
- red spaces = itinerary and unordered list

Interface:
- left side = itinerary
- right side = unordered
- map in back
Screen Shot 2020-06-24 at 5 02 42 PM

Add item pop up - event form
Screen Shot 2020-06-24 at 5 02 54 PM

Add item pop up - flight form
Screen Shot 2020-06-24 at 5 03 02 PM

Date time picker
Screen Shot 2020-06-24 at 4 53 59 PM
Screen Shot 2020-06-24 at 4 53 51 PM


step-capstone/src/Components/AddEventHotel.js
```diff
@@ -0,0 +1,89 @@
...
+            </Grid>
+            <Grid item>
+                <Checkbox
+                    checked={checked}
+                    onChange={handleChecked}
+                />
+            </Grid>
+            <Grid item container direction="row" justify="space-between">
+                <MuiPickersUtilsProvider utils={DateFnsUtils}>
+                    <DateTimePicker value={startDate} onChange={handleStartChange} />
```
ganewman42, line 65: Labeling the start and end dates explicitly would be helpful. For people that sometimes read right to left instead of left to right the fact that the left one is the start time might not be obvious. This would be good for all of the start/end time forms.

step-capstone/src/Components/AddFlight.js
```diff
@@ -0,0 +1,93 @@
...
+    MuiPickersUtilsProvider,
+    DateTimePicker
+} from '@material-ui/pickers';
+import DateFnsUtils from "@date-io/date-fns";
+
+export default function AddFlight(props) {
+    const [departureDate, handleStartChange] = useState(new Date());
+    const [arrivalDate, handleEndChange] = useState(new Date());
+    const [checked, setChecked] = useState(false);
+    const [departureAirport, setDepartureAirport] = useState("");
```
ganewman42, line 18: Something to think about: Do you care about whether these are airport codes or place names? Do you want to mention that in the form or validate it somewhere?

step-capstone/src/components/AddFormPopover.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import {Popover} from '@material-ui/core';
+import AddForm from './AddForm'
+
+export default function AddFormPopver(props) {
```
ganewman42, line 5: Curious: How well does this popover perform on a variety of window and screen sizes?

step-capstone/src/components/AddItemButton.js
```diff
@@ -0,0 +1,43 @@
...
+import AddFormPopover from './AddFormPopover'
+
+export default function AddItemButton(props) {
+    const [anchorEl, setAnchorEl] = React.useState(null);
+
+    const handleClick = (event) => {
+        setAnchorEl(event.currentTarget);
+    };
+
+    const handleClose = () => {
```
ganewman42, line 13: Does this mean that once the popover is open, it cannot be closed? If not, how can you close it?

step-capstone/src/App.js
```diff
@@ -1,12 +1,46 @@
...
-    return(
+  constructor(props) {
+    super(props);
+    this.state = {
+      data: []
+    }
+  }
+
+  fetchData() {
+    return [
```
ganewman42, line 15: Could this be wired up to the trip dataset you define below instead? That seems like better test data.

step-capstone/src/components/Trip.js
```diff
@@ -0,0 +1,247 @@
...
+            // Add to database here
+            // current code for testing data format etc.
+            switch (type) {
+                case "event":
+                    console.log("adding event");
+                    break;
+                case "hotel":
+                    this.setState({
+                        items: this.state.items.concat([{
+                            id: this.state.items.length,
```
ganewman42, line 149: How does this guarantee uniqueness, not only within a logged in user, but between users?

step-capstone/src/components/Trip.js
```diff
@@ -0,0 +1,247 @@
...
+            console.log("please enter information")
+        } else {
+            // Add to database here
+            // current code for testing data format etc.
+            switch (type) {
+                case "event":
+                    console.log("adding event");
+                    break;
+                case "hotel":
+                    this.setState({
```
ganewman42, line 147: This re-paramaterizing seems a little redundant. Why can't you simply pass the values, or validate them and then pass?

step-capstone/src/components/Trip.js
```diff
@@ -0,0 +1,247 @@
...
+                            type: type,
+                            departureAirport: data.departureAirport,
+                            arrivalAirport: data.arrivalAirport,
+                            departureDate: data.departureDate.toString(),
+                            arrivalDate: data.arrivalDate.toString(),
+                            description: data.description
+                        }])
+                    })
+                    break;
+                default: console.log("invalid type"); break;
```
ganewman42, line 175: If this happens, then your code has done something very unexpected. you should probably do something more severe than logging, at least a warning and possibly an error.

step-capstone/src/components/AddFormPopover.js
emaohn: it's position and sizing will properly adjust itself if the window size changes. If it becomes too small it will become scrollable.

step-capstone/src/components/AddItemButton.js
emaohn: You can close it either by clicking out outside of it or either pressing cancel or save

step-capstone/src/components/Trip.js
emaohn: it doesn't--this was just for testing. With real data each item should have it's own unique key/id

emaohn: I kept it in for the prototype

step-capstone/src/App.js
emaohn: Oh this is really old code, sorry. I've moved everything to Trip.js instead of App but I didn't push App

step-capstone/src/components/Trip.js
emaohn: I originally did it to just add the type attribute to the object, but you're right there must be a less redundant way to do that. But when actually adding to the database we'd have to do something like this for each property of the item.

emaohn:
Updates: 
- Add editing feature: Click edit button to pull up information and modify it
- Additional labeling: labeling for inputs to make clear intentions to user
- "Error validation": Ensures mandatory inputs are filled out before saving --> if not, pressing save will not close the popover and non-optional inputs will be red

Screen Shot 2020-06-24 at 11 31 01 PM
Screen Shot 2020-06-24 at 11 31 11 PM
Screen Shot 2020-06-24 at 11 31 17 PM
Screen Shot 2020-06-24 at 11 31 36 PM

step-capstone/src/components/Flight.js
```diff
@@ -18,8 +18,8 @@ export default function Flight( data ) {
     return(
         <Grid item>
             <Typography variant="h6" gutterBottom>Flight from { data.departureAirport } to { data.arrivalAirport }</Typography>
-            <Typography variant="subtitle2" gutterBottom>Departure: { data.departureDate }</Typography>
-            <Typography variant="subtitle2" gutterBottom>Arrival: { data.arrivalDate }</Typography>
+            <Typography variant="subtitle2" gutterBottom>Departure: { data.departureDate.toString() }</Typography>
+            <Typography variant="subtitle2" gutterBottom>Arrival: { data.arrivalDate.toString() }</Typography>
```
zirving, line 22: I can't remember, but I thought we decided we would handle already be passing string versions of dates into our components?

emaohn: Yeah, so I realized that trip object actually needs to have the real date representation in order to be able to load the edit popover to display the previously selected date in the date input

step-capstone/src/App.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import './styles/App.css';
 import TravelObject from './components/TravelObject'
+import { Box, Grid, Typography } from '@material-ui/core'
 
 class App extends React.Component {
```
hieudan225, line 5: Should we rename this as Trip?

hieudan225: Oh I see your change

step-capstone/src/components/Trip.js
```diff
@@ -1,18 +1,23 @@
...
+import { Grid } from '@material-ui/core'
 
 export default class Trip extends React.Component {
     constructor(props) {
         super(props);
         this.state = {
             data: []
         }
+
+        this.deleteObject = this.deleteObject.bind(this);
```
hieudan225: Just a question out of curiosity, what does this line of code mean, Emmie?

hieudan225: I guess its purpose is to pass down the function the sub component: Travel Object? I still wonder what "bind" does.

emaohn: Basically the "this" keyword doesn't work as it does it java as we're used to, and binding "this" to the function ensures that anytime this is called (for example, passing down this.function), it's referring to this class component. A common work around is using arrow functions, but I think "this binding" is generally preferred.

step-capstone/src/components/TravelObject.js
```diff
@@ -22,20 +22,24 @@ export default function TravelObject(props) {
...
                     </Grid>
                     <Grid item>
                         <Grid container direction="column">
-                            <Edit
-                                onClick={() => props.onEditItem(props.data.id)}
-                            />
-                            <Delete
-                                onClick={() => props.onRemoveItem(props.data.id)}
-                            />
+                            <Grid item>
```
hieudan225, line 35: Just a question to confirm my understanding: to add scrolling functionality to itinerary container, we need all the travel object to be in grid? or what makes the scrolling functionality possible?

emaohn: I have it so that scrolling will automatically be enabled in the itinerary container if travelobjects continue past the given space (it's in the css)

step-capstone/src/components/AddItemButton.js
hieudan225: Another question out of curiosity, how did you implement the feature of "closing the popup by clicking outside of it"?

step-capstone/src/components/Trip.js
emaohn: Learn more here: https://reactjs.org/docs/faq-functions.html

step-capstone/src/components/AddItemButton.js
emaohn: I believe that simply the characteristic of the Popover component provided by Material UI: https://material-ui.com/components/popover/#popover

I didn't have to implement that part. The only thing I did have to do was tell it to close when pressing cancel or save

step-capstone/src/components/Trip.js
cthamrun: If this is temporary please add a comment to specify a TODO to replace it with something that guarantees uniqueness like Grace mentioned.

PR #7, opened 2020-06-24
Firebase set up and authentication

hieudan225:
Firebase is set up in src/components/Firebase.js.

There is only firebase instance created in the whole app (has been created in index.js). Please use that instance instead of creating a new one to prevent unwanted behavior.

To use firebase instance, make sure that your component is wrapped in FirebaseContext.Provider (index.js) and then use existing firebase instance by calling "this.context" in your class component. (One example is app.js) There is  other way to access the context in a function component as well, please refer to React Context API.

Firebase authentication has now worked, except for the Signing in with 3rd party organization feature. Please use signing in with Email for now. You can access to more user's information by modifying the getUserInfo in firebase.class. See the example of how to do this in app.js.


ganewman42:
This PR includes all of the node modules, which should definitely not be the case. You should add node_modules to your gitignore or otherwise make sure they don't get committed.

hieudan225:
@ganewman42 oh sorry, I forget about that. I hope you can tell me more on what to do now. Should I delete the yarn.lock file?

ganewman42:
No, the lock file should be fine. Probably your best bet would be to add `node_modules/` to your .gitignore

step-capstone/src/App.js
```diff
@@ -1,27 +1,55 @@
...
-  SomeComponent = () => (
-    <FirebaseContext.Consumer>
-      {(firebase) => {
-        firebase.creatFirebaseWidget();
-      }}
-    </FirebaseContext.Consumer>
-  );
-  value = this.SomeComponent();
+  afterAuthStateCheck(status) {
+    console.log(status);
```
emaohn: maybe remove this?

step-capstone/src/App.js
```diff
@@ -1,27 +1,55 @@
...
+    console.log(this.context);
+    this.context.getUserInfo().then( (status) => this.afterAuthStateCheck(status));
+
+  }
   render () {
-    return(
-      <div className="App">
-        <p> Hello World </p>
-        {this.SomeComponent()}
+    console.log(this.state);
```
emaohn: more debugging statements to be removed?

step-capstone/src/App.js
```diff
@@ -1,27 +1,55 @@
...
+    const display = (this.state.authState)? "block" : "none";
+    console.log(display);
```
emaohn: same ^^

step-capstone/src/App.js
```diff
@@ -1,27 +1,55 @@
...
+    const displayObject = {display: display};
+    return (
+      <div>
+          <div id="firebaseui-auth-container"></div>
```
emaohn: Since the App component acts as a router to different components (ex. login, home, etc) the login/logout UI should probably be their own components that the app can route to for specific URLS (and we'd set the URL depending on login state as you've written above).

step-capstone/public/index.html
```diff
@@ -6,28 +6,24 @@
...
+    <link type="text/css" rel="stylesheet" href="https://www.gstatic.com/firebasejs/ui/4.5.1/firebase-ui-auth.css" />
+  
   </head>
   <body>
-    <noscript>You need to enable JavaScript to run this app.</noscript>
-    <div id="root"></div>
-      <div id="firebaseui-auth-container"></div>
+    <a class="brand-logo">
+      <img src="images/zed-logo.png" style="width: 100px; margin-top: 10px;">
+      <h1>Welcome to Zed - The Smart Travel Assistant</h1>
```
emaohn: This would end up showing on every page I think, but we probably wouldn't want that with the user's home page and the trip interface

PR #8, opened 2020-06-24
App

zirving:
- added react-router-dom : handles loading components for different url endings

- added App skeleton: not finished, but will be finished once we figure out authentication, etc.


step-capstone/src/App.js
```diff
@@ -1,15 +1,60 @@
...
+          </Switch>
+        </Router>
+      
       </div>
     );
   }
 }
+function Login(){
+  return ( 
+    <Login/>
```
emaohn, line 44: Would rendering itself cause some error?

zirving: probably that was just a placeholder till I get Dan's authentication stuff. happened to use the same name

PR #9, opened 2020-06-25
Trip item component

emaohn:
**NOTE: Since I branched off Trip component, all commits from there transferred here. Only last 4 commits are relevant to this PR**
Screen Shot 2020-06-25 at 5 05 10 PM

- Setup trip card that can be used to display all the trips on the home screen
- setup routing function between home screen and individual trips (commented out, but can be uncommented when merged with the routing branch from Zach)

Screen Shot 2020-06-25 at 3 19 25 PM


step-capstone/src/App.js
```diff
@@ -1,12 +1,49 @@
...
+    };
+
+    this.handleOpenTrip = this.handleOpenTrip.bind(this);
+  }
+
+  handleOpenTrip(tripID) {
+    // fetch trip from database with tripID
+    this.setState({ selectedTrip: "fetched trip" })
+    console.log("Opening trip " + tripID);
+    // redirect to trip component
```
ganewman42, line 21: I think this is something you intend to do later. For comments like that the convention is to put `TODO ({your username}): {whatever it is you'll do later}`. That makes it easy to search through the code for stuff you need to finish and for your teammates to understand what's up. You do this in a few places but not consistently.

step-capstone/src/Components/FormPopover.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import {Popover} from '@material-ui/core';
+import ItemForm from './ItemForm'
+
+export default function FormPopver(props) {
```
ganewman42, line 5: Nice, I love a good utility function. If you think you'll make similar decisions in a few places (eg `onClose` will probably always have the same behavior), consider reducing the amount of things in props and instead taking a small number of arguments for what you really care about.

step-capstone/src/components/AddEventHotel.js
```diff
@@ -0,0 +1,165 @@
...
+                <TextField
+                    error
+                    helperText="Cannot leave field blank"
+                    id="title"
+                    label={"Add Title"}
+                    defaultValue={title}
+                    fullWidth
+                    onChange={handleTitleChange}
+                />
+            )
```
ganewman42, line 81: This, together with the above, seems like a lot of code to handle a single validation. Rather than writing basically the same hook for a few different fields, think about whether you can make a common utility so you only do it once.

step-capstone/src/components/ItemForm.js
```diff
@@ -0,0 +1,136 @@
...
+                    type="event"
+                    data={this.state.data}
+                    onToggleValidation={this.handleToggleValidation}
+                />
+            )
+        } else if ((!isNew && this.state.data.type === "flight") || (isNew && this.state.value === 1)) {
+            return (
+                <AddFlight
+                    onDataChange={this.handleDataChange}
+                    data={this.state.data}
```
ganewman42, line 68: Is `type="flight"` missing?

step-capstone/src/components/TripItemComponent.js
```diff
@@ -0,0 +1,44 @@
...
+const useStyles = makeStyles({
+    root: {
+        maxWidth: 345,
+    },
+    media: {
+        height: 140
+    },
+})
+/*
+* Trip component should recieve some trip
```
ganewman42, line 22: Nit: Spelling, receive

step-capstone/src/Components/FormPopover.js
emaohn: Could you elaborate on this? Do you mean condensing repeated pass downs into singular objects to reduce clutter, for example, putting onClose, addItem, and onEdit into a singular functions object? I still need onClose to be passed down so I can't necessarily just not pass it into props.

step-capstone/src/components/ItemForm.js
emaohn: It seemed unnecessary to pass down the type since in creating an AddFlight component the type must be flight. I passed down the type into event hotel since it could either be an event or a hotel.

step-capstone/src/components/AddEventHotel.js
emaohn: Good point! I'll try to find a way.

step-capstone/src/Components/FormPopover.js
ganewman42: What I was trying to say is that if you will always be passing the same value for some options in `props`, it might be easier to pass a few individual arguments rather than a large `props` structure with a lot of options. However, it seems like this file is not in the PR anymore so resolving.

step-capstone/src/components/AddEventHotel.js
```diff
@@ -65,62 +65,18 @@ export default function AddEventHotel(props) {
...
-    // render error input if no title
-    let renderTitleInput = () => {
-        if (title === "") {
-            return (
+    return (
+        <Grid container direction="column">
+            <Grid item>
                 <TextField
-                    error
+                    error={title === ""}
```
ganewman42, line 72: Nice, this validation pattern is a lot simpler.

PR #12, opened 2020-06-25
Firebase set up and authentication

hieudan225:
Firebase is set up in src/components/Firebase.js.

To use firebase instance, make sure that your component is wrapped in FirebaseContext.Provider (index.js) and then use existing firebase instance by calling "this.context" in your class component. (One example is app.js) There is other way to access the context in a function component as well, please refer to React Context API.

Firebase authentication has now worked, except for the Signing in with 3rd party organization feature. Please use signing in with Email for now. You can access to more user's information by modifying the getUserInfo in firebase.class. See the example of how to do this in app.js.


step-capstone/public/index.html
```diff
@@ -1,43 +1,30 @@
...
-
-      You can add webfonts, meta tags, or analytics to this file.
-      The build step will place the bundled scripts into the <body> tag.
-
-      To begin the development, run `npm start` or `yarn start`.
-      To create a production bundle, use `npm run build` or `yarn build`.
-    -->
+    <a class="brand-logo">
+      <img src="images/zed-logo.png" style="width: 100px; margin-top: 10px;">
+      <h1>Welcome to Zed - The Smart Travel Assistant</h1>
```
emaohn, line 24: I think I mentioned this in your previous PR, but this is going to show up the entire time with the App component under it, which we probably don't want when the user is signed in and viewing the full screen map/trip component/dashboard?

emaohn: It might be best to have this displaying only on the login component rather than all the time

step-capstone/src/App.js
```diff
@@ -1,15 +1,68 @@
...
-    return(
-      <div className="App">
-        <TravelObject type="flight" />
+    console.log(this.state);
+    const display = (this.state.authState)? "block" : "none";
+    console.log(display);
+    const displayObject = {display: display};
+    return (
+      <div>
+          <div id="firebaseui-auth-container"></div>
```
emaohn, line 52: Should be its own login component?

step-capstone/src/App.js
```diff
@@ -1,15 +1,68 @@
...
+          
+          <div id="sign-in-status"></div>
+          <div id="sign-in">
```
emaohn, line 55: Same as above, but we should probably have a logout component as well

step-capstone/public/index.html
hieudan225: Yes Emmie, I have just made the SignInWidget component class that includes the brand, logo and sign-in buttons.

step-capstone/src/App.js
hieudan225: This comment has been addressed: SignOutButton has been created! ^^

step-capstone/package.json
```diff
@@ -9,6 +9,7 @@
     "@testing-library/react": "^9.3.2",
     "@testing-library/user-event": "^7.1.2",
     "react": "^16.13.1",
+    "react-bootstrap": "^1.0.1",
```
zirving, line 12: why are we adding bootstrap? I though we were using material-UI as our styling

hieudan225:
@googlebot I signed it!

step-capstone/package.json
hieudan225: You're right, since I just want to add a button in fast and I mistakenly wrote "Button" instead of "button", the HTML tag didn't recognize that, I went to stackoverflow and add react-bootstrap in just to test the sign-out button. I can remove it now.

PR #13, opened 2020-06-26
Create Event Class

hieudan225:
Create Event Class

### DIFF
--- a/step-capstone/src/App.js
+++ b/step-capstone/src/App.js
@@ -6,7 +6,7 @@ class App extends React.Component {
   render () {
     return(
       <div className="App">
-        <TravelObject type="flight" />
+        <TravelObject type="event" data={data} />
       </div>
     );
   }
--- a/step-capstone/src/components/Event.js
+++ b/step-capstone/src/components/Event.js
@@ -1,0 +1,27 @@
+import React from 'react'
+import { Box, Grid, Typography } from '@material-ui/core'
+
+/*
+Takes data in this form:
+        {
+          finalized: true,
+          type: "event",
+          title: "Visiting Sherlock",
+          location: "221b Baker Street, London",
+          timeStart: "8:30",
+          timeEnd: "10:30",
+          description: "Additional notes"
+        }
+
+*/
+export default function Flight( data ) {
+    return(
+        <Grid item>
+            <Typography variant="h6" gutterBottom>Title { data.title } to { data.title }</Typography>
+            <Typography variant="subtitle2" gutterBottom>Location: { data.location }</Typography>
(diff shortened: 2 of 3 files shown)


step-capstone/src/App.js
```diff
@@ -6,7 +6,7 @@ class App extends React.Component {
   render () {
     return(
       <div className="App">
-        <TravelObject type="flight" />
+        <TravelObject type="event" data={data} />
       </div>
     );
   }
```
zirving, line 12: could you pull this branch down and git checkout this file? Its gonna definitely cause a merge conflict with App and doesnt need to be committed

step-capstone/src/components/Event.js
```diff
@@ -0,0 +1,27 @@
...
+          type: "event",
+          title: "Visiting Sherlock",
+          location: "221b Baker Street, London",
+          timeStart: "8:30",
+          timeEnd: "10:30",
+          description: "Additional notes"
+        }
+
+*/
+export default function Flight( data ) {
```
zirving, line 17: change Flight to Event

step-capstone/src/components/Event.js
```diff
@@ -0,0 +1,27 @@
...
+    return(
+        <Grid item>
+            <Typography variant="h6" gutterBottom>Title { data.title } to { data.title }</Typography>
+            <Typography variant="subtitle2" gutterBottom>Location: { data.location }</Typography>
+            <Typography variant="subtitle2" gutterBottom>Time start: { data.timeStart }</Typography>
+            <Typography variant="subtitle2" gutterBottom>Time end: { data.timeEnd }</Typography>
```
zirving, line 23: I dont think we want all the descriptions of what each thing is to appear. so it should just be {data.location} for example, not Location: {data.location}. Also, I think Title is still left in a data format, and Time Start and Time End, if we do display them, should be done like start - end

step-capstone/src/components/Event.js
```diff
@@ -0,0 +1,27 @@
...
+import { Box, Grid, Typography } from '@material-ui/core'
+
+/*
+Takes data in this form:
+        {
+          finalized: true,
+          type: "event",
+          title: "Visiting Sherlock",
+          location: "221b Baker Street, London",
+          timeStart: "8:30",
```
ganewman42, line 11: Timezones would be a good thing to add here, including display handling for them.

step-capstone/src/components/Event.js
ganewman42: I agree with not labeling title and location, but labeling start and end times makes an equal amount of sense to me as start-end if you prefer that.

emaohn: I believe the date/time will come in as a Data object, so you would need to convert that to a string format (toString())

PR #15, opened 2020-07-01
Map visualizations

emaohn:
- Add markers to map for all travel objects with a specified location 
- Draws path between finalized travel objects (solid line between flights and dotted line between everything else)
- Color coordinated per travel object type
- Map will automatically zoom to encompass all markers upon render
- Click on map marker to zoom into location
- Changed Map to MapComponent (interfering with Map class in javascript)

Screen Shot 2020-07-01 at 01 07 10

Screen Shot 2020-07-01 at 01 05 50

### DIFF
--- a/step-capstone/src/components/Map.js
+++ b/step-capstone/src/components/Map.js
@@ -49,7 +49,7 @@ class MapComponent extends React.Component {
     })
   }
 
-  addMarker(coordinates, type) {
+  addMarker(coordinates, type, label) {
     // TODO: firebase will provide coordinates as a GeoPoint --> convert to {lat, lng}
     // { lat: coordinates.lat(), lng: coordinates.lng() }
 
@@ -90,17 +90,15 @@ class MapComponent extends React.Component {
     };
 
     let geoPath = new window.google.maps.Polyline({
-      path: path.path,
+      path: path,
       strokeColor: '#FF0000',
-      strokeOpacity: path.type === "normal" ? 0.0 : 1.0,
-      icons: path.type === "normal"
-        ? [{
+      strokeOpacity: 0,
+      strokeWeight: 2,
+      icons: [{
           icon: lineSymbol,
           offset: '0',
           repeat: '20px'
         }]
-        : [],
-      strokeWeight: 2,
     });
 
     geoPath.setMap(map);
@@ -136,7 +134,7 @@ class MapComponent extends React.Component {
    *  Draws path between all finalized travel objects, returns list of path objects
    */
   drawPaths() {
-    // Separate into different segments: flight vs non-flight
+    // Separate into different segments
     var paths = [];
     var curPath = [];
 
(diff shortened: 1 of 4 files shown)


ganewman42:
Some comments on the design that are optional depending on your thoughts:

* I'm not a huge fan of the lines for the flight, since your flight route is not a route you follow in the same way you go other places, I'm worried it could just be visually confusing. Consider leaving it off or styling it differently to make it clearer. For example, flight visualizations often use arcs instead of lines. 
* It would be nice if there was an indicator of directionality or timing on the places you're visiting. Some ideas:
  * Put numbers on the pins indicating the visit order
  * Add arrow heads to the connecting lines to show the direction
  * Add colors or shading to the lines or pins to show order (darker colors are earlier in the day or similar)

hieudan225:
Could we display a real path between the places? I mean the real paths may be more useful to users than straight lines?

emaohn:
- Removed path between flights
- Shade of path to get darker as it gets later into the day

step-capstone/src/components/Map.js
```diff
@@ -6,37 +6,217 @@ import React, { createRef } from 'react'
...
+  addPath(map, path, color) {
+    // TODO : change to show directions between paths
+    // dotted line
+    var lineSymbol = {
+      path: 'M 0,-1 0,1',
+      strokeOpacity: 1,
+      scale: 4
+    };
+
+    console.log(color);
```
zirving, line 93: leftover from testing

step-capstone/src/components/Map.js
```diff
@@ -6,37 +6,215 @@ import React, { createRef } from 'react'
...
+        bounds.extend(item.arrivalCoordinates);
+      }
+      return hashMap;
+    }, new Map())
+  }
+
+  /*
+   *  Draws path between all finalized travel objects, returns list of path objects
+   */
+  drawPaths() {
```
ganewman42, line 137: This function ended up being quite long. I would consider separating the color helper into a little separate function to help with the length.

step-capstone/src/components/Trip.js
```diff
@@ -12,99 +12,85 @@ const testData = [
...
-        finalized: true,
-        type: "flight",
-        departureAirport: "BOS",
-        arrivalAirport: "SFO",
-        departureDate: new Date(),
-        arrivalDate: new Date(),
-        description: "Additional notes"
+        id: 327,
+        finalized: false,
+        type: "hotel",
```
ganewman42, line 45: Shouldn't this be an event, not a hotel? Is event handling not ready yet? Applies to the below as well.

step-capstone/src/components/Map.js
```diff
@@ -6,37 +6,219 @@ import React, { createRef } from 'react'
...
+      geoPaths: []
+    }
+    this.googleMapRef = createRef();
+    this.googleMap = undefined
+  }
 
-class Map extends React.Component {
-  
   componentDidMount() {
+    console.log("MOUNTING");
```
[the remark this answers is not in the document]
emaohn: I think this was already removed?

step-capstone/src/components/Map.js
```diff
@@ -6,37 +6,219 @@ import React, { createRef } from 'react'
...
+      animation: window.google.maps.Animation.DROP,
+      icon: { url: iconUrl }
+    })
+
+    // zoom to marker when clicked
+    newMarker.addListener('click', () => {
+      this.googleMap.setZoom(15);
+      this.googleMap.setCenter(newMarker.getPosition());
+
+      // TODO: this.props.onMarkerClicked(id) --> notifies parent that marker clicked
```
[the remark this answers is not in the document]
emaohn: I was thinking when the marker is clicked, the travelobject is scrolled to / highlighted in the itinerary. Needed later because it is not implemented yet / not yet merged with itinerary component. I put it there as a placeholder when it would may eventually be needed.

step-capstone/src/components/Trip.js
emaohn: I think Event component hasn't been merged yet. Having an event object is fine, but won't do anything right now so I only tested with hotels which have the exact same properties as events.

PR #16, opened 2020-07-01
Firestore Database Implementation

zirving:
- firebase.js: Adds the ability to read data from the database, using either getData() or getCollection() as well as add, edit and delete travel objects. there are also addTrip and deleteTrip methods, however there is currently no functionality in TripList to use these. All these methods return a Promise when completed - see Trip and TripList for examples on how to handle. I also added notes to the class and class structure doc on how to use them and what exactly "reference" is. 

- App.js: added saving of userId to sessionStorage when logged in. added signoutButton. 

- TripList.js: componentDidmount: pulls data from database according to the user Id and saves it to the displayed trips[] array. also handles the user selection of a trip. 

- Trip.js: componentDidMount: pulls travelObjects from database according to user and trip Ids, and adds them to the displayed items[] array. the handleAdd, Edit, and Delete methods handle their respective functions. Currently, they do not force a reloading of all the data from the database. instead, they change both the database and the local items[] array. this is to keep our site speedy. 

Minor Changes: 
package.json/yarn.lock - added necessary dependencies. 
Hotel.js - location is no longer passed in as a string, so conversion needs to happen here now. 
TripItemComponent - data is now all passed in through data field, to seperate tripId and onClick methods. 
TravelObject -returning the object rather than its id is easier for delete to handle.

### DIFF
--- a/step-capstone/src/App.js
+++ b/step-capstone/src/App.js
@@ -33,7 +33,6 @@ class App extends React.Component {
 
     if (status.signInStatus) {
       this.setState({authState: true, user: status.user});
-      // this.setState({authState: true, user: status.user, reference: "/users/"+this.context.auth.currentUser.uid});
       console.log("You have logged in.");
     }
     else {
--- a/step-capstone/src/App.js
+++ b/step-capstone/src/App.js
@@ -10,6 +10,7 @@ import {FirebaseContext} from './components/Firebase';
 import SignInWidget from './components/Firebase/SignInWidget';
 import SignOutButton from './components/Firebase/SignOutButton';
 import TripList from "./components/TripList"
+import Trip from "./components/Trip"
 
 class App extends React.Component {
   static contextType = FirebaseContext;
@@ -19,33 +20,20 @@ class App extends React.Component {
     
     this.state = {
       authState: null,
-      user: null
+      user: null,
     };
 
   }
 
   componentDidMount() {
-    
     this.context.getUserInfo().then( (status) => this.afterAuthStateCheck(status));
-    //Example code: Read the db function
-    // this.context.db.collection('users').get().then((snapshot) => {
-    //   snapshot.docs.forEach(doc => {
-    //     console.log(doc.data());
-    //   })
-    // });
-
-    // //Example code: Write to the db functionif(this.state.authState){
-    // this.context.db.collection('users').add({
-    //   displayName: "memo",
-    //   email: "memo@email.com"
-    // });
-
   }
 
   afterAuthStateCheck(status) {
 
     if (status.signInStatus) {
       this.setState({authState: true, user: status.user});
+      // this.setState({authState: true, user: status.user, reference: "/users/"+this.context.auth.currentUser.uid});
       console.log("You have logged in.");
     }
     else {
@@ -55,8 +43,10 @@ class App extends React.Component {
   }
 
   handleLogin(){
-    console.log("handleLogin");
     if(this.state.authState){
+      //TODO: replace the hardcoded test user with logged in user once all users can create trips
+      //sessionStorage.setItem("reference", "/users/"+this.context.auth.currentUser.uid);
+      sessionStorage.setItem("userId", "0fmXVWHePsZoCV6ex1Z2");
       return <Redirect to = "/trips/"/>;
     } else {
       return <Redirect to = "/login-page"/>;
@@ -64,17 +54,21 @@ class App extends React.Component {
   }
 
   render () {
+   
     return(
       <div className="App">
         {this.state.authState ? <SignOutButton/> :''}
         <Router>
           {this.handleLogin()}
           <Switch>
-            <Route path="/login-page">
+            <Route exact path="/login-page">
               <Login />
             </Route>
-            <Route path = "/trips/">
-              <TripList userId = 'userId'/>
+            <Route path = "/trips/:tripId">
+              <Trip/>
+            </Route>
+            <Route exact path = "/trips/">
+              <TripList />
             </Route>
           </Switch>
         </Router>
--- a/step-capstone/src/components/Firebase/firebase.js
+++ b/step-capstone/src/components/Firebase/firebase.js
@@ -56,13 +56,13 @@ class Firebase {
   }
 
   getTrip(reference){
-    const doc = this.db.doc(reference);
-    return doc.get();
+    const tripRef = this.db.doc(reference);
+    return tripRef.get();
   }
 
   getTripList(reference){
-    const collec = this.db.collection(reference);
-    return collec.get();
+    const tripListRef = this.db.collection(reference);
+    return tripListRef.get();
   }
 
   addTrip(reference, trip) {
@@ -85,10 +85,10 @@ class Firebase {
      })
     });
   }
-
-  editTrip(reference, data) {
-    //TODO: Implement when details of trip editing and settings are implemented
-  }
+// TODO: Implement when details of trip editing and settings are implemented
+//   editTrip(reference, data) {
+    
+//   }
 
   deleteTravelObject(reference, travelObject) {
     const tripRef = this.db.doc(reference);
--- a/step-capstone/src/components/Firebase/firebase.js
+++ b/step-capstone/src/components/Firebase/firebase.js
@@ -55,12 +55,12 @@ class Firebase {
     });
   }
 
-  getData(reference){
+  getTrip(reference){
     const doc = this.db.doc(reference);
     return doc.get();
   }
 
-  getCollection(reference){
+  getTripList(reference){
     const collec = this.db.collection(reference);
     return collec.get();
   }
--- a/step-capstone/src/components/Firebase/firebase.js
+++ b/step-capstone/src/components/Firebase/firebase.js
@@ -55,6 +55,51 @@ class Firebase {
     });
   }
 
+  getData(reference){
+    const doc = this.db.doc(reference);
+    return doc.get();
+  }
+
+  getCollection(reference){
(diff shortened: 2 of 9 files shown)


step-capstone/src/components/Firebase/firebase.js
```diff
@@ -55,6 +55,51 @@ class Firebase {
     });
   }
 
+  getData(reference){
```
ganewman42, line 58: This collection of functions is a little confusing because there's no overriding organizational scheme. You're going to end up with a lot of these and consistency will make it easier to follow. What I might recommend is having a get, add, edit and delete for each type of object with the same naming pattern. For example, `getTravel`, `addTravel`, `editTravel`, `deleteTravel`. These could then call private helper methods as needed.

Alternatively, it might make sense to put common helpers here like `getCollection` and `getData`, then do the actual implementations for each object in that object's class.

step-capstone/src/components/Trip.js
```diff
@@ -112,78 +21,59 @@ export default class Trip extends React.Component {
...
-        })
+      let travelObjectList = [];
+      this.context.getData(this.state.reference)
+      .then(data => {
+        data.data().travelObjects.forEach(travelObject => {
+          travelObjectList.push(travelObject)
+        });
+        this.setState({items : travelObjectList});
+      })
+      .catch(error => {console.log("Error Getting Trip Data")});
```
ganewman42, line 33: If you're logging an error, make sure to include the original error message as well or it will be hard to debug. Let me know if react does this automatically but I couldn't find documentation of that. This applies to pretty much all your error handling.

step-capstone/src/components/Trip.js
```diff
@@ -112,78 +21,59 @@ export default class Trip extends React.Component {
...
+              if (item.id === data.id) {
+                let success = this.context.editTravelObject(this.state.reference, item, data)
+                  .then(() => {
+                    return true;
+                  })
+                  .catch(error => {
+                    console.log("Error Editing Item");
+                    return false; 
+                  });
+                return (success ? data : item);
```
emaohn, line 58: Is it guaranteed that the editing has been complete at this point? It seems possible that success might be undefined here. I thought only things happening in the .then() were guaranteed to happen when the process is finished, though I could be wrong.

In that case, you could probably eliminate the success variable and instead of returning a boolean, return data when complete and item if error occurs.

step-capstone/src/components/Trip.js
```diff
@@ -112,78 +21,59 @@ export default class Trip extends React.Component {
...
+                  return item;
+              }
+          })
+      })
     }
 
     handleAddItem(data) {
         if (data === undefined) {
-            console.log("please enter information")
+            console.log("please enter information");
```
emaohn, line 68: Should probably throw an error if this happens rather than a console.log(), because if it does something went wrong with the form (it's not supposed to allow creation of empty travel objects)

step-capstone/src/components/Trip.js
zirving: you are right. i am not entirely sure how to do this correctly - originally I had it the way you are talking about, but that wasnt working either, as anything returned within a then or catch seems to be in the wrong scope - when I return from there it does not return to the .map function. I think I can rework it to use a forEach instead and then just set the item equal to the promise chain like I do with success here but open to ideas here.

step-capstone/src/components/Trip.js
zirving: i agree though I am not sure who would catch that error? and would we just want to display the form again?

step-capstone/src/App.js
```diff
@@ -19,33 +20,20 @@ class App extends React.Component {
...
-    //   email: "memo@email.com"
-    // });
-
   }
 
   afterAuthStateCheck(status) {
 
     if (status.signInStatus) {
       this.setState({authState: true, user: status.user});
+      // this.setState({authState: true, user: status.user, reference: "/users/"+this.context.auth.currentUser.uid});
```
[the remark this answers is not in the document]
zirving: needed to be removed, thanks

step-capstone/src/components/Firebase/firebase.js
zirving: good idea, I will switch to using a common naming scheme.

step-capstone/src/components/Trip.js
zirving: done, thanks

step-capstone/src/components/Trip.js
```diff
@@ -50,23 +50,44 @@ export default class Trip extends React.Component {
...
           })
-      })
+          .catch((error) => {
+            console.log("Error Editing Item");
+            console.log(error);
+          });
+      console.log(newItems.length);
+      this.setState({items : newItems});
+      console.log(this.state.items);
+      // this.setState({
```
emaohn, line 74: maybe remove this commented out section

step-capstone/src/components/Trip.js
hieudan225: I don't think we'll encounter this error at all, since we'll only call handleAddItem when the user click Save and the handleSave() function will first check if the data is validated (if the data is empty or the dates are wrong...). Then, only when the data is validated, will the date be saved and the form popover be closed.
image

step-capstone/src/components/Trip.js
hieudan225: I think `forEach` and `map` are kind of similar. Current code looks alright to me.

step-capstone/src/components/TripList.js
```diff
@@ -1,60 +1,53 @@
...
-    const trips = this.getTrips();
+    const trips = this.state.trips;
     return (
       <div>
-        <Switch>
-          <Route path = "/trips/:tripId">
-            <Trip tripId = {this.state.selectedTrip}/>
-          </Route>
-        </Switch>
         {this.state.selectedTrip ? <Redirect to = {"/trips/"+this.state.selectedTrip}/> :
```
hieudan225, line 48: My current thought is redirecting the trip can be done just using onClick function? Why do we need to use State for this?

zirving: one reason to have it like this so in the future, if we wanted to be able to get refresh working, we could check if a trip was selected, even if nobody had actually clicked the button. basically, it makes it so the criteria for loading a trip is whether that trip has been selected, rather than if the user specifically clicked the button triggering onClick.

step-capstone/src/components/Trip.js
zirving: ok I agree. I will just remove that section

step-capstone/src/components/TripList.js
emaohn: I believe it had to do with the whole refresh issue zach was having. Basically when refreshing it would try to go to the trip URL without knowing which one was clicked on since it was essentially "restarting" so we need to store locally the trip the user is currently on to go back.

step-capstone/src/components/Firebase/firebase.js
```diff
@@ -55,6 +55,51 @@ class Firebase {
     });
   }
 
+  getTrip(reference){
```
ganewman42, line 58: Nice, I like the new organizational scheme for these functions.

step-capstone/src/components/Firebase/firebase.js
```diff
@@ -55,6 +55,51 @@ class Firebase {
...
+      .then(() => {
+        tripRef.update({travelObjects : firebase.firestore.FieldValue.arrayUnion(newTravelObject)})
+        .then(() => {resolve()});
+     })
+    });
+  }
+
+  editTrip(reference, data) {
+    //TODO: Implement when details of trip editing and settings are implemented
+  }
```
ganewman42, line 91: Nit: To avoid accidentally calling a function that does nothing, you might want to comment out and TODO the whole thing.

step-capstone/src/components/Trip.js
```diff
@@ -112,80 +21,71 @@ export default class Trip extends React.Component {
...
+      this.context.getTrip(this.state.reference)
+      .then(data => {
+        data.data().travelObjects.forEach(travelObject => {
+          travelObjectList.push(travelObject)
+        });
+        this.setState({items : travelObjectList});
+      })
+      .catch(error => {
+        console.log("Error Getting Trip Data")
+        console.error(error)
```
ganewman42, line 35: Logging the error seems like a good approach. Even better would be surfacing an error back to the user, but I think this makes sense for the MVP.

step-capstone/src/components/Firebase/firebase.js
zirving: done

step-capstone/src/components/Trip.js
zirving: yes I agree. all the error handling right now is pretty weak

step-capstone/src/components/Firebase/firebase.js
```diff
@@ -55,6 +55,51 @@ class Firebase {
     });
   }
 
+  getTrip(reference){
+    const doc = this.db.doc(reference);
+    return doc.get();
+  }
+
+  getTripList(reference){
+    const collec = this.db.collection(reference);
```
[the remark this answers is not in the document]
zirving: also changed

step-capstone/src/components/Firebase/firebase.js
```diff
@@ -55,6 +55,51 @@ class Firebase {
     });
   }
 
+  getTrip(reference){
+    const doc = this.db.doc(reference);
+    return doc.get();
```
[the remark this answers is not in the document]
zirving: good catch, thanks, it is changed

PR #17, opened 2020-07-06
Finalized and Unfinalized 

hieudan225:
Core features added in this pull request:
- Finalized and Unfinalized rendering
- Trip Setting Popover (functionality implemented: reset dateStart will change displayDate)
- Display Date Picker and Arrows (finished functionality)
- Dropdown Menu button (haven't added routing yet)
- Validate the Dates input in Adding Flight, Event, Hotel and new Item

Future features:
- changing the center of the map to the selected destinations
- asking users if they want to delete a certain date if they change the dateStart and dateEnd
- TimeZone button! : select TimeZone for display


hieudan225:
image

hieudan225:
You can try new features of this pull request here: 
https://privacy-engprod-capstone.ue.r.appspot.com/

hieudan225:
image
Trip Setting Popover

hieudan225:
image
Drop down Menu (haven't added routing yet)

hieudan225:
image
Date Picker Calendar and Arrows

step-capstone/src/components/Event.js
```diff
@@ -0,0 +1,24 @@
...
+         finalized: true,
+         type: "event",
+         title: "Visiting Sherlock",
+         location: "221b Baker Street, London",
+         startDate: "8:30",
+         endDate: "10:30",
+         description: "Additional notes"
+       }
+*/
+export default function Flight( props ) {
```
zirving, line 15: typo: Flight should be Event

step-capstone/src/components/Event.js
```diff
@@ -0,0 +1,24 @@
...
+   return(
+       <div className="event double events" style={props.styleConfig}>
+           <p>{props.data.startDate.toLocaleString()} - {props.data.endDate.toLocaleString()}: {props.data.title}</p>
```
zirving, line 18: At least to me, it would be more readable to have the Title first, then a seperate <p> for the date afterwards. Ditto on flight and hotel.

zirving: haha didnt realize this would convert my html, a seperate paragraph for the date.

step-capstone/src/components/Finalized.js
```diff
@@ -0,0 +1,132 @@
...
+    const [displayDate, setDisplayDate] = useState(props.startDate);
+
+    const handleDisplayDateChange = (newDisplayDate) => {
+        setDisplayDate(newDisplayDate);
+    }
+
+    const sameDate = (timeA, timeB) => {
+      return (timeA.getDate() === timeB.getDate() && timeA.getMonth() === timeB.getMonth() && timeA.getFullYear() === timeB.getFullYear());
+    }
+    const pushItemToArrayInHashMap = (item, dateKey) => {
```
zirving, line 18: not super pressing but the array in hashmap naming is not super helpful. i want to know what the hashmap and array are in relation to the code (hashmap is our dates, array is the list of items on the date) as I already know by their declaration what data structure they are. could you try renaming the hashmap, array and related and methods for clarity.

step-capstone/src/components/ItemForm.js
```diff
@@ -111,22 +111,22 @@ export default class ItemForm extends React.Component {
...
-            <Card id="add-form-container">
+            <Card id="add-form-container" style={{height: "350px"}}>
                 <CardContent>
                     {this.renderTab()}
                     {this.getForm()}
                 </CardContent>
                 <CardActions>
                     <Button onClick={this.state.onClose} size="small">Cancel</Button>
                     <Button onClick={this.handleSave} size="small">Save</Button>
+                    <Button onClick={() => this.props.onRemoveItem(this.state.data.id)} size="small">Delete</Button>
```
zirving, line 129: why do we need to be able to delete the item form?

step-capstone/src/components/Unfinalized.js
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import TravelObject from './TravelObject';
+import { Grid } from '@material-ui/core';
+import TripSettingPopover from "./TripSettingPopover";
+
+function Header(props) {
+
+    function dropDownFunc(e) {
```
zirving, line 8: like it has a form popover that we are using, material UI also has a Menu https://material-ui.com/components/menus/ - something to consider

step-capstone/src/index.js
```diff
@@ -2,11 +2,12 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import './styles/index.css';
 import App from './App';
+import Trip from './components/Trip';
 import * as serviceWorker from './serviceWorker';
 
 ReactDOM.render(
   <React.StrictMode>
-    <App />
+    <Trip />
```
zirving, line 10: why is this no longer app?

step-capstone/src/components/ItemForm.js
zirving: oh i see. this is the same form that appears when you click on the event. not sure how difficult it would be to change, but now, when you click the plus button, it also has a delete button, which doesnt do anything.

ganewman42:
FYI this page does not display as expected for me: 
image

step-capstone/src/components/AddEvent.js
```diff
@@ -1,60 +1,72 @@
...
-
-export default function AddEventHotel(props) {
+export default function AddEvent(props) {
     let overwriting = props.data !== undefined;
-
     // Sets values to previous values if editing, otherwise blank slate
-    const [startDate, handleStartChange] = useState(overwriting ? props.data.startDate : new Date());
-    const [endDate, handleEndChange] = useState(overwriting ? props.data.endDate : new Date());
+    const [startDate, setStartDate] = useState(overwriting ? props.data.startDate : props.startDate);
+    const [endDate, setEndDate] = useState(overwriting ? props.data.endDate : props.startDate);
```
ganewman42, line 18: This looks odd since you use start date here as well, should it be `const [endDate, setEndDate] = useState(overwriting ? props.data.endDate : props.endDate);`? This applies to the file below as well.

step-capstone/src/components/Finalized.js
```diff
@@ -0,0 +1,132 @@
...
+    })
+    return (
+        <TimeLine 
+            displayDate = {displayDate}
+            onChangeDisplayDate = {handleDisplayDateChange}
+        />
+    )
+}
+ 
+
```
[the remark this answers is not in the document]
hieudan225: Just like we discussed before, for Hotel, we won't be displaying them across multiple days, but we'll have CheckIn and CheckOut event?
Each event will have its own startDate and endDate (which should be named precisely as startTime and endTime, but considering the uniform naming that we are using, I still kept it startDate and endDate here).
Thank you Zach for noticing this confusion, I suggest that we will revise the Hotel Object to contain:
Check in:
       Start Time: <DateTimePicker>
       End Time: <DateTimePicker>
Check out:
       Start Time: <DateTimePicker>
       End Time: <DateTimePicker>

then display these two Checkin and Checkout event seperately in Finalized?
What do you think?

step-capstone/src/components/AddEvent.js
```diff
@@ -1,60 +1,72 @@
...
+
+    }
+    const handleEndDateChange = (newEndDate) => {
+        if (startDate >= newEndDate) {
+            var newStartDate = new Date(newEndDate);
+            newStartDate.setTime(newEndDate.getTime() - 5*60000);
+            setStartDate(newStartDate);
+        }
+        setEndDate(newEndDate);
+    }
```
ganewman42, line 53: These functions confuse me a little. It looks like their purpose is to update the start or end point to be a fixed unit of time from whichever point was just set. Could you clarify why these are here? Also, what are the time units on the offset you use?

step-capstone/src/components/ItemForm.js
hieudan225: The functionality of this Delete button is actually to delete the Travel Object.

step-capstone/src/components/Unfinalized.js
hieudan225: You're right. Let me change that

step-capstone/src/index.js
hieudan225: I put Trip here just for faster debugging. I'll put it back to App.

hieudan225:
@ganewman42 yes Grace, since I just want to show the itinerary, I didn't change the Map API to a deployed one I guess, since I don't do anything with Map in this pull request.

step-capstone/src/components/Finalized.js
```diff
@@ -0,0 +1,132 @@
...
+        if (startDate.getMinutes() < 30) {
+            return output + ":00";
+        }
+        else {
+            return output + ":30";
+        }
+    }
+    const getHeightPercentage = (startDate, endDate) => {
+        let dif = endDate.getHours()*60 + endDate.getMinutes() - startDate.getHours()*60 - startDate.getMinutes();
+        return dif*48.0/30-16;
```
ganewman42, line 73: The values here (especially the 48) seem like a result of tuning. For such values, it can be wise to create a global variable rather than set value throughout your code. That way, if you ever need to change it, you can change it in one place, and it's clear what the value does up front.

step-capstone/src/components/TripSettingFormPopover.js
```diff
@@ -0,0 +1,171 @@
...
+    handleValidation(isValidated) {
+        if (this.state.isValidated !== isValidated) {
+            this.setState({isValidated : isValidated});
+        }
+    }
+    handleDataChange(newTripSetting) {
+        this.setState({tripSetting: newTripSetting});
+    }
+    handleSave() {
+        // Doesn't save data if improper input --> doesn't close popover which shows error
```
ganewman42, line 36: Curious: How does the error get shown?

step-capstone/src/components/AddHotel.js
```diff
@@ -0,0 +1,186 @@
...
+                   defaultValue={title}
+                   fullWidth
+                   onChange={handleTitleChange}
+               />
+           )
+       }
+   }
+ 
+   // Render error input if finalized item has no location
+   let renderLocationInput = () => {
```
emaohn, line 114: Instead of having two cases for two different text fields, you can merge it into one (without this render function) that sets error = null and helperText = "" if no error exists.

step-capstone/src/components/AddHotel.js
```diff
@@ -0,0 +1,186 @@
...
+           props.onToggleValidation(false);
+       } else if (checked && location === "") {
+           props.onToggleValidation(false);
+       } else {
+           props.onToggleValidation(true);
+       }
+   }, [startDate, endDate, checked, title, location, description])
+ 
+   // render error input if no title
+   let renderTitleInput = () => {
```
emaohn, line 87: Also don't need two different cases as mentioned below --> can do directly in render function

step-capstone/src/components/AddItemButton.js
```diff
@@ -16,14 +16,14 @@ export default function AddItemButton(props) {
...
     const open = Boolean(anchorEl);
-
     return (
         <div>
             <Fab aria-label="add" onClick={handleClick}>
                 <AddIcon />
             </Fab>
             <FormPopover
                 data={undefined}
+                startDate={props.startDate}
```
emaohn, line 26: Why do we need to pass in a startDate into the form?

step-capstone/src/components/Finalized.js
```diff
@@ -0,0 +1,132 @@
...
+      return (timeA.getDate() === timeB.getDate() && timeA.getMonth() === timeB.getMonth() && timeA.getFullYear() === timeB.getFullYear());
+    }
+    const pushItemToArrayInHashMap = (item, dateKey) => {
+        if (!hashMap.has(dateKey)) {
+            hashMap.set(dateKey, [item]);
+        }
+        else {
+            let array = hashMap.get(dateKey);
+            array.push(item);
+            hashMap.set(dateKey, array);
```
emaohn, line 25: you probably don't need to reset the array into the hashmap since it's still referencing the same array

step-capstone/src/components/Finalized.js
emaohn: This method could be named "addItemToDate()"

step-capstone/src/components/Finalized.js
```diff
@@ -0,0 +1,132 @@
...
+        else if (travelObjectA.startDate >  travelObjectB.startDate) {
+            return 1;
+        }
+        else   {
+            return -1;
+        }
+    }
+    const sortHashMap = () => {
+        const iterator = hashMap[Symbol.iterator]();
+        for (let KV of iterator) {
```
emaohn, line 52: What are KV and V? Maybe make variable names more clear?

step-capstone/src/components/Finalized.js
```diff
@@ -0,0 +1,132 @@
...
+            return 0;
+        }
+        else if (travelObjectA.startDate >  travelObjectB.startDate) {
+            return 1;
+        }
+        else   {
+            return -1;
+        }
+    }
+    const sortHashMap = () => {
```
emaohn, line 50: From my understanding, this function sorts the arrays inside the HashMap, but my initial thought when reading the function name was that the entire map itself was going to be sorted by date or something--perhaps you could either add comments for clarity or rename (or both)?

step-capstone/src/components/Finalized.js
```diff
@@ -0,0 +1,132 @@
...
+            let V = KV[1];
+            V.sort(travelObjectstartDateComparator);
+        }
+    }
+    
+    useEffect(() => {
+        setDisplayDate(props.startDate);
+    }, [props.startDate]);
+    
+    const getAssociatedDiv = (startDate) => {
```
emaohn, line 62: I'm a bit confused as to what this does/what it's for?

emaohn: is it its place in the itinerary?

step-capstone/src/components/Finalized.js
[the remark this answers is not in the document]
emaohn: I think that's a good idea @hieudan225

step-capstone/src/components/Flight.js
```diff
@@ -1,26 +1,25 @@
...
-            <Typography variant="subtitle2" gutterBottom>Arrival: { data.arrivalDate.toString() }</Typography>
-            <Typography variant="body2" gutterBottom>{ data.description }</Typography>
-        </Grid>
-    )
-}
\ No newline at end of file
+export default function Flight( props ) {
+   return(
+       <div className="event double flights" style={props.styleConfig}>
+           <p>{props.data.startDate.toLocaleString()} - {props.data.endDate.toLocaleString()}: Flight from { props.data.departureAirport } to { props.data.arrivalAirport }</p>
```
emaohn, line 20: I think this has potential to be better formatted: for example, having some sort of header (preferably depart/arrival airports), and under, the the corresponding times. It does take a bit more space but I think the one line of sparation could add a good amount to readability and UX. Also since the boxes are so small, chances are the text will be overflowed into the next line anyways.
Also, just for consistence, I think we should probably also stick with using Typography from Material UI for all our text (they do have a paragraph text option though I think we should bold or make bigger the airports since p tags are very small).

emaohn: Same with others (event, hotel)

step-capstone/src/components/ItemForm.js
emaohn: I agree with @zirving in that we should think of the case when we're adding an object, in which we would have no reason to present the user with a delete button. Maybe only render this button when we're editing?

step-capstone/src/components/TimeLine.js
```diff
@@ -0,0 +1,166 @@
...
+    setDisplayDate(props.displayDate);
+  }, [props.displayDate])
+
+  useEffect(() => {
+    props.onChangeDisplayDate(displayDate);
+  }, [displayDate])
+
+  return (
+    <MuiPickersUtilsProvider utils={DateFnsUtils} id="timeline">
+      <KeyboardDatePicker style={{
```
emaohn, line 23: you could move style into the next line so the indentation isn't so out of place

step-capstone/src/components/TimeLine.js
```diff
@@ -0,0 +1,166 @@
...
+  }
+  return (
+    <div className="icon secondary chevron_left" style={{left: "15px"}} onClick={handleOnClick}>›</div>
+  )
+}
+
+
+function Header(props) {
+  return (
+    <div>
```
emaohn, line 65: Something to keep in mind perhaps for later changes: Using material UI components for UI consistency in all our components. For example, small things like button and text.

step-capstone/src/components/TripSettingFormPopover.js
```diff
@@ -0,0 +1,171 @@
...
+            description :  description
+        })
+        // notifies form if necessary inputs are present
+        props.onValidation(!(destinations === ""))
+    }, [destinations, startDate, endDate, description])
+    // Users must input destinations --> renders error input if not
+    let renderDestinations = () => {
+        if (destinations === "") {
+            return (
+                <TextField
```
emaohn, line 106: similar to other forms, don't need the cases and can do in one

step-capstone/src/components/AddEvent.js
hieudan225: To clarify, the variable `props.startDate` is the overall trip's startDate. When user adds a new item (the `props.data` will be undefined), the AddEvent (same for AddFlight and AddHotel) will initialized the Date to be the startDate of the trips instead. I think it is more natural for a event to end in one day, so I initialized both startDate and endDate equal to the overall trip's startDate.

step-capstone/src/components/AddEvent.js
hieudan225: To clarify, these two functions ensure that `startDate` must be smaller than `endDate` (their names is Date but actually they are referring to Time; first I started with `timeStart` and `timeEnd`; but I refactored it to fit Emmie and Zach's code).
For example, If the user changes the time of the `endDate` to a time that is smaller than the current `startDate`, this function will change the `startDate` to be at least 5 minutes before the new `endDate`. Let me know if this explanation still confuses you.

step-capstone/src/components/AddHotel.js
hieudan225: Good suggest for more concise code! I will make changes!

step-capstone/src/components/AddItemButton.js
hieudan225: I pass the startDate to initialize the startDate of any new TravelObject. Before, we'll initialize the date as the current date (new Date(), now they are props.startDate). You can see the use of this in Add[TravelObject] files.

step-capstone/src/components/Event.js
hieudan225: woops, you are right. Thank you for noticing this!

step-capstone/src/components/Event.js
hieudan225: Good point! I will change the order of the rendering!

step-capstone/src/components/Finalized.js
hieudan225: Good suggestions! I will make sure to change the names of the method and data structures!

step-capstone/src/components/Finalized.js
hieudan225: you're right. Abundant line of code.

step-capstone/src/components/Finalized.js
hieudan225: Good point! I will rename the function and add comments to clarify that,

step-capstone/src/components/Finalized.js
hieudan225: Yes, this function gets the div that the TravelObject clings to.

hieudan225: Let me add comments to the function to clarify

step-capstone/src/components/Finalized.js
hieudan225: That's a good suggestion. I will make sure to put these const as a global variable.

step-capstone/src/components/Flight.js
hieudan225: Good point.

step-capstone/src/components/ItemForm.js
hieudan225: Ah, turns out that's what Zach means. Okay, I will make changes so that the button only renders when editing.

step-capstone/src/components/TimeLine.js
hieudan225: Yes!

step-capstone/src/components/TimeLine.js
hieudan225: Ok, I'll remember that

step-capstone/src/components/TripSettingFormPopover.js
hieudan225: image
It looks like this

step-capstone/src/components/AddEvent.js
cthamrun: Instead of doing this implicit behavior, would it instead be possible to instead throw an error to the user to have them set the endDate to a time later than the startDate?

step-capstone/src/components/TripSettingFormPopover.js
cthamrun: There seems to be a typo in the screenshot ("Departire" instead of "Departure").

hieudan225: Thank you for noticing that!

step-capstone/src/components/AddEvent.js
hieudan225: In the current code, I try to replicate the behavior of Google Calendar, but yes, we can also throw the error to user. What do you think?

cthamrun: From what I tried in Calendar, it seems that Calendar prevents you from creating an even that has endTime prior to startTime in the first place. I'm not sure which behavior you're referring to (maybe I'm misunderstanding) but I still prefer not allowing the user to create such an event in the first place (refuse to take input for event creation). Maybe @ganewman42 has a different opinion here.

hieudan225: @cthamrun I think you have misunderstood my point. So the current behavior happens during the editting process. The user will has no chance to submit a TravelObject if that has invalidated time range, because the invalidated time range will be reset to be validated again. So the current behavior is: during the editing process, if the new endDate is smaller than the current startDate, the current startDate will be set to be 5 minutes before the new endDate; if the new startDate is bigger than the current endDate, the current endDate will be set to be 5 minutes after the new startDate.

hieudan225: Screenshot 2020-07-06 at 7 27 04 PM
For example, in Google Calendar, I have an event from 5:30 to 6

hieudan225: If I change the start time to 6:15, it will automatically change the end time to be 6:45 
Screenshot 2020-07-06 at 7 27 18 PM

hieudan225:
image
This is the latest look:

step-capstone/src/components/Flight.js
hieudan225: done

step-capstone/src/components/Event.js
hieudan225: done

step-capstone/src/components/Finalized.js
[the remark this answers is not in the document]
hieudan225: * Note: this task will be implemented next week. The Hotel object will also be displayed as a side bar.

step-capstone/src/components/ItemForm.js
hieudan225: Done

step-capstone/src/components/Unfinalized.js
hieudan225: Update: Since the other buttons are configured in raw CSS styles, now when I change the drop down button using Material UI, it costs so much time to regain the styles. Thus, I consider not to change it.

hieudan225:
I have made a lot more changes to Finalized.js. It would be great if you could review that file!

step-capstone/src/components/TimeLine.js
```diff
@@ -0,0 +1,167 @@
...
+                  <OneHourInterval idV="14" />
+                  <OneHourInterval idV="15" />
+                  <OneHourInterval idV="16" />
+                  <OneHourInterval idV="17" />
+                  <OneHourInterval idV="18" />
+                  <OneHourInterval idV="19" />
+                  <OneHourInterval idV="20" />
+                  <OneHourInterval idV="21" />
+                  <OneHourInterval idV="22" />
+                  <OneHourInterval idV="23" />
```
ganewman42, line 153: Not being a front end person I wonder if this could be generated, but you're the expert now so up to you.

step-capstone/src/components/Trip.js
```diff
@@ -1,256 +1,231 @@
...
+              endDate : new Date(),
+              description : ""
+          }
+      }
+      this.handleRemoveItem = this.handleRemoveItem.bind(this);
+      this.handleEditItem = this.handleEditItem.bind(this);
+      this.handleAddItem = this.handleAddItem.bind(this);
+      this.handleEditTripSetting = this.handleEditTripSetting.bind(this);
+  }
+  fetchData() {
```
ganewman42, line 95: Nit: It's nice to have extra newlines between function definitions, maybe put those back and use elsewhere.

step-capstone/src/components/Unfinalized.js
ganewman42: I would say it should at least be a separate PR, up to you all whether you think the change is worth the effort period.

step-capstone/src/components/AddFlight.js
```diff
@@ -1,121 +1,109 @@
...
+    This behavior happends during editting, so the user can never submit an invalid time range. 
+    */
+    const handleEndDateChange = (newEndDate) => {
+        if (startDate >= newEndDate) {
+            var newStartDate = new Date(newEndDate);
+            newStartDate.setTime(newEndDate.getTime() - endDate.getTime() + startDate.getTime());
+            setStartDate(newStartDate);
+        }
+        setEndDate(newEndDate);
+    }
```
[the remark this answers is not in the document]
hieudan225: Addressed

step-capstone/src/components/Flight.js
```diff
@@ -1,26 +1,37 @@
...
-        </Grid>
-    )
-}
\ No newline at end of file
+export default function Flight( props ) {
+  let timeString = props.data.startDate.toLocaleString() + ' - ' + props.data.endDate.toLocaleString();
+  /*Given 2 Date objects, return true if they have the same date; return false otherwise */
+  const sameDate = (timeA, timeB) => {
+    return (timeA.getDate() === timeB.getDate() && timeA.getMonth() === timeB.getMonth() && timeA.getFullYear() === timeB.getFullYear());
+  }
```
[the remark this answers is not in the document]
hieudan225: It's a cool idea thou.

step-capstone/src/components/TimeLine.js
hieudan225: That's a good idea. I will put a for loop in to generate this.

hieudan225: Addressed

step-capstone/src/index.js
ganewman42: This still needs to be fixed.

hieudan225:
@ganewman42 oh, did I miss any comments?

ganewman42:
@hieudan225 Yes, this one: https://github.com/googleinterns/step98-2020/pull/17/files#r451140528 where you have `Trip` instead of `App`

hieudan225:
@ganewman42 oh I see, is it the one from Zach?

PR #21, opened 2020-07-09
Hotel interval fix + Fix bug involving multiple events within an hour timespan

emaohn:
- Hotel checkin and checkout renders as two separate 30 minute time blocks on their respective days
- Fixed bug that didn't allow more than 1 travel object to be rendered within a one hour time interval (for example, wouldn't allow an event from 1:00-1:15 and 1:30-1:50) and caused following events to not get rendered

### DIFF
--- a/step-capstone/src/components/TimeLine.js
+++ b/step-capstone/src/components/TimeLine.js
@@ -28,11 +28,14 @@ function OneHourInterval(props) {
 
     } else if (sameDate(props.displayDate, startDate)) {
       dif = 24 * 60 - startDate.getHours() * 60 - startDate.getMinutes();
-
     } else {
       dif = endDate.getHours() * 60 + endDate.getMinutes();
     }
     return dif * displayHeightOfADiv / minPerDiv - padding;
+  }
+
+  const getHeightFromMin = (min) => {
+    return min * displayHeightOfADiv / minPerDiv - padding;
   }
 
   /*Given a startDate of a TravelObject, return the top pixel*/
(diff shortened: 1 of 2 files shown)


step-capstone/src/components/TimeLine.js
```diff
@@ -139,22 +139,22 @@ export default function TimeLine(props) {
   }
 
   /* Handling rendering starts HERE */
-
   separateDates();
   sortItemList();
 
   var displayItems = (date2Items.has(displayDate.toDateString())) ? date2Items.get(displayDate.toDateString()) : [];
   var nextItemIndex = 0;
+  console.log("display items length: ", displayItems.length)
```
zirving, line 147: leftover from debugging

step-capstone/src/components/TimeLine.js
```diff
@@ -139,22 +139,22 @@ export default function TimeLine(props) {
...
   for (var i = 0; i < 24; i++) {
     if (nextItemIndex < displayItems.length) {
       var nextItem = displayItems[nextItemIndex];
       var nextItemStartDate = nextItem.startDate;
-      if (!sameDate(displayDate, nextItemStartDate)) {
+      // End date within current display date 
+      if (!sameDate(displayDate, nextItemStartDate) && (nextItem.type !== "hotel" || nextItem.endDate.getHour === i)) {
```
zirving, line 153: what does this if handle? hotels and ?

step-capstone/src/components/TimeLine.js
```diff
@@ -178,13 +177,13 @@ export default function TimeLine(props) {
             div = ":30";
```
zirving, line 177: confused how this works? shouldn't the div be :00 if less than 30? and where does div get set to :00 currently?

zirving: oh nevermind git was randomly hiding those lines for me very strange

PR #22, opened 2020-07-09
Trip Adding Button, Trip Setting, Firebase methods for deleting and editing trip

hieudan225:
1. Changes in `Add[TravelObject]`: to fix props.data undefined bug. The reason for the existence of these added useEffect() hook is:
`ItemForm` actually controls the `startDate` and `endDate`, so `Add[TravelObject] `needs to listen to props.data.[start/end]Date to update its state.

2. Add Trip adding button, which will call `TripSettingPopover` when clicked. If user saves the form, new trip will be added to database.

3. Add Trip deleting button, which will delete trip from database.

4. Edit Trip setting method, which will take information of `TripSettingFormPopover` to edit trip setting (any fields of trip except for `travelObjects`) in database.


step-capstone/src/components/AddEvent.js
```diff
@@ -34,6 +34,18 @@ export default function AddEvent(props) {
         setDescription(e.target.value);
     }
 
+    useEffect(() => {
```
emaohn, line 37: Maybe put these in the same useEffect since they both happen at the same time?

emaohn: Also I thought we took this out? From my understanding, useEffect is used when component states are changed. If props.data changes the entire function would be re-rendered anyways since it would be as state change in the parent component (form) which automatically triggers a re-render of this component and any other Add component.

step-capstone/src/components/AddFlight.js
```diff
@@ -35,12 +35,17 @@ export default function AddFlight(props) {
     }
     
     useEffect(() => {
```
emaohn, line 37: Same as above

step-capstone/src/components/AddHotel.js
```diff
@@ -40,13 +40,18 @@ export default function AddHotel(props) {
        setDescription(e.target.value);
    }
    
-    useEffect(() => {
```
emaohn, line 43: Same as above

step-capstone/src/components/Firebase/firebase.js
```diff
@@ -69,6 +72,51 @@ class Firebase {
     return tripListRef.add(trip);
   }
 
+  async deleteTrip(reference) {
+    const tripRef = this.db.doc(reference);
+    return new Promise((resolve, reject) =>
+     {
+      tripRef.delete().then(() => {
+        console.log("Trip successfully deleted!");
```
emaohn, line 80: Should we remove the console.logs? Theres some others in this file too, not sure if that was intentional or if you forgot to remove them

step-capstone/src/components/TripList.js
```diff
@@ -33,30 +38,64 @@ class TripList extends React.Component{
...
   }
 
+
   handleOpenTrip(tripId) {
     this.setState({selectedTrip : tripId})
     sessionStorage.setItem("tripId",tripId);
   }
 
+  handleAddTrip(newTrip) {
+    console.log("Adding new trip to database...");
```
emaohn, line 56: remove print?

step-capstone/src/components/TripList.js
```diff
@@ -33,30 +38,64 @@ class TripList extends React.Component{
...
+    const reference = "/users/" + this.state.userId + "/trips";
+    this.context.addTrip(reference, newTrip);
+    this.loadTrips();
+  }
+
+  handleDeleteTrip(tripId) {
+    const reference = "/users/" + this.state.userId + "/trips/" + tripId;
+    console.log(reference);
```
emaohn, line 64: Same as above

step-capstone/src/components/Firebase/firebase.js
```diff
@@ -69,6 +72,51 @@ class Firebase {
...
+        resolve();
+      }).catch(function(error) {
+          console.error("Error removing trip: ", error);
+          reject()
+      })
```
zirving, line 85: I think you can remove this promise here. if you just return tripRef.delete() it will return a promise to its caller. Then then and catch should be handled by the caller, because this firebase function does not know what needs to happen after the delete succeeds or fails.

step-capstone/src/components/Trip.js
```diff
@@ -16,23 +16,35 @@ export default class Trip extends React.Component {
             reference: "/users/" + sessionStorage.getItem("userId") + "/trips/" + sessionStorage.getItem("tripId"),
             items: [],
             tripSetting: {
+                title: "",
                 startDate: new Date(),
                 endDate: new Date(),
-                destinations: []
+                destinations: "",
```
zirving, line 22: why change destinations to a string? it seems more like an array of strings to me

step-capstone/src/components/TripList.js
```diff
@@ -33,30 +38,64 @@ class TripList extends React.Component{
     .catch(error => {
       console.log("Error Getting Trips")
       console.error(error)
-    });
+    })
+  }
+
+  componentDidMount() {
+    this.loadTrips();
```
zirving, line 45: this is probably unnesscary to have a method where all it does is call another method

zirving: oh I see why you did this, more explanation later

step-capstone/src/components/TripList.js
```diff
@@ -33,30 +38,64 @@ class TripList extends React.Component{
...
   handleOpenTrip(tripId) {
     this.setState({selectedTrip : tripId})
     sessionStorage.setItem("tripId",tripId);
   }
 
+  handleAddTrip(newTrip) {
+    console.log("Adding new trip to database...");
+    const reference = "/users/" + this.state.userId + "/trips";
+    this.context.addTrip(reference, newTrip);
+    this.loadTrips();
```
zirving, line 59: this is something we should probably all discuss, but here and in delete you reload trips from firebase after an addition or a deletion. which works, but how we handled things in Trip was we change the local version of trips[], so that we dont have to reload and call to firebase on every change. not as big of a deal here but might be nice to decide on one approach

step-capstone/src/components/TripList.js
```diff
@@ -33,30 +38,64 @@ class TripList extends React.Component{
...
-                  <TripItemComponent key={trip.id} data={trip.data()} tripId={trip.id} onOpenTrip={this.handleOpenTrip} />
-                </Grid>
-              );
-            })
-          }
+          <Grid id="trips" 
+            container spacing = {2} 
+            style={{position: "absolute", 
+                    top: "100px", 
+                    left: "100px"}}
```
zirving, line 77: maybe it would be better to make these a %? for different screen sizes

step-capstone/src/components/TripList.js
```diff
@@ -33,30 +38,64 @@ class TripList extends React.Component{
...
+                      onOpenTrip ={this.handleOpenTrip}
+                      onDeleteTrip = {this.handleDeleteTrip} />
+                  </Grid>
+                );
+              })
+            }
+                  <Grid item>
+                    <AddTrip
+                      onAddTrip = {this.handleAddTrip}/>
+                  </Grid>
```
zirving, line 95: unindennt a couple i think should be inline with trips.map

step-capstone/src/components/TripSettingFormPopover.js
```diff
@@ -91,62 +94,74 @@ function EditTripSetting(props) {
...
+            title: title,
             destinations : destinations,
             startDate : startDate,
             endDate : endDate,
-            description :  description
+            description :  description,
         })
         // notifies form if necessary inputs are present
-        props.onValidation(!(destinations === ""))
+        props.onValidation(!(destinations === "" || (title === "")))
```
zirving, line 104: do we need them to list destinations?

step-capstone/src/components/AddEvent.js
hieudan225: Hi Emmie, so this is the short answer for why: https://stackoverflow.com/questions/38892672/react-why-child-component-doesnt-update-when-prop-changes#:~:text=Because%20children%20do%20not%20rerender,if%20its%20STATE%20changes%20%3A)&text=It%20will%20pass%20data%20from,is%20no%20rerender%20logic%20there.&text=According%20to%20React%20philosophy%20component,parent%20and%20should%20be%20immutable.
Here is one experiment from our app:
image

You could see that in the left, ItemForm rerender with 3AM - 6AM (start - end), and the new props in AddEvent also shows 3 AM - 6 AM. But the form, in reality, doesn't update the state. That's why I added these useEffect to listen to the change and update the state.

step-capstone/src/components/Firebase/firebase.js
hieudan225: Oh, this one is intentional? I thought this was similar to "You have logged in" and sort of thing? Should I delete it?

step-capstone/src/components/TripList.js
hieudan225: This should be removed! Thank you!

step-capstone/src/components/Firebase/firebase.js
hieudan225: good point

step-capstone/src/components/Trip.js
hieudan225: it should be an array, but since in the scope of this PR, to sync with the current test data of firebase database and the current code, I changed it to a string just to test things out. However, Emmie's part will ultimately remove this and replace this with her autocomplete which will handle the destinations. So yes, this change will be removed when we come to merge the code.

step-capstone/src/components/TripList.js
hieudan225: componentDidMount() ensures the order of implementation: which is after things are being rendered, it will load the trips. So I thought it has some meaning there? What do you think?

step-capstone/src/components/TripSettingFormPopover.js
hieudan225: as I said above, this part will be handled by Emmie's code.

step-capstone/src/components/TripList.js
hieudan225: Yes, so I was trying to put AddTrip in the same indentation as TripItemComponent. Do you think it's a bad idea?

step-capstone/src/components/Firebase/firebase.js
hieudan225: addressed

step-capstone/src/components/TripList.js
hieudan225: good point

step-capstone/src/components/TripList.js
hieudan225: addressed

step-capstone/src/components/TripList.js
hieudan225: Updated (After Zach and I discussed): To avoid reload the whole things, we need to manually give the trip its ID. We can do that, but it requires changing a lot of thing. Probably we can do so in the different PR.

step-capstone/src/components/AddFlight.js
hieudan225: addressed above.

step-capstone/src/components/AddHotel.js
hieudan225: addressed above.

PR #23, opened 2020-07-09
New Homepage

zirving:
- fixed some minor issues with firebase double load and tripitemcomponent date display

- created new Homepage!


hieudan225:
Also LGTM

PR #27, opened 2020-07-10
Fix UI bug in MVP

hieudan225:
1. Fix OneHourInterval's inconsistent indentation
2. Fix trip setting's endDate error
3. Fix Travel Object's height
4. Add border to Travel Object

### DIFF
--- a/step-capstone/src/styles/TimeLine.css
+++ b/step-capstone/src/styles/TimeLine.css
@@ -103,6 +103,14 @@
   tr, tr td {
     height: 20px;
   }
+  .OneHourInterval .Interval {
+    position: absolute;
+    left: 55px;
+    width: 228px;
+  }
+  .OneHourInterval .headcol {
+    width: 40px;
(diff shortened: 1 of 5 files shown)


step-capstone/src/styles/TimeLine.css
```diff
@@ -1,3 +1,21 @@
...
+  align-items: center;
+  align-content: center;
+  justify-content: center;
+  font-family: "DINPro", "Helvetica Neue", sans-serif;
+  margin: 0;
+  background: #fafafa;
+  box-sizing: border-box;
+  height: 100vh;
+
+}
```
zirving, line 18: this shouldn't be added back right?

hieudan225: woop, that should be deleted. Thank you Zach!

hieudan225: addressed

PR #28, opened 2020-07-10
Additional map interactions

emaohn:
- Zoom to marker when travel object is clicked in itinerary
- Draw paths only between events under selected date in itinerary
- Zoom in/out to encompass all of selected days' events when toggling between days in itinerary

### DIFF
--- a/step-capstone/src/components/Map.js
+++ b/step-capstone/src/components/Map.js
@@ -29,14 +29,19 @@ class MapComponent extends React.Component {
    * Updates state with any changes in props
    */
   componentDidUpdate(prevProps) {
-    if (this.props.selected !== null) {
-      let selectedMarker = this.state.finalizedMarkers.has(this.props.selected)
-        ? this.state.finalizedMarkers.get(this.props.selected).marker
-        : this.state.unfinalizedMarkers.get(this.props.selected).marker;
+    console.log(this.props.today.events)
+    if (this.props.selected !== null && this.props.selected !== prevProps.selected) {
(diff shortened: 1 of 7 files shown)


step-capstone/src/components/Map.js
```diff
@@ -108,14 +128,15 @@ class MapComponent extends React.Component {
     // construct hashmap with key: travelObject id, value: marker object
     return list.reduce((hashMap, item) => {
```
hieudan225, line 129: maybe, we could name the hashMap as `travelObject2Marker`?

step-capstone/src/components/Map.js
```diff
@@ -108,14 +128,15 @@ class MapComponent extends React.Component {
...
       if (item.type !== "flight") {
-        hashMap.set(item.id, { marker: this.addMarker(item.coordinates, item.type, item.id), type: item.type });
+        hashMap.set(item.id, { marker: this.addMarker(item.coordinates, item.type), type: item.type });
```
hieudan225, line 131: maybe, we could name the hashMap `travelObject2Marker`?

step-capstone/src/components/Trip.js
```diff
@@ -21,35 +21,45 @@ export default class Trip extends React.Component {
                 endDate: new Date(),
                 destinations: "",
                 description: ""
+            },
+            loaded: false,
+            selectedObject: null,
+            today: {
```
hieudan225, line 27: maybe consider naming `today` as `displayDate` for consistency

emaohn: good suggestion! will do :)

PR #33, opened 2020-07-16
Suggestion tests

zirving:
This branch contains our written tests and some changes to our methods so that they pass. 

Three test files: CountCat.js, which tests the algorithim to loop through a user's preferences and the fetched results and count how many categories match. GetScore.test.js which tests the scoring metric. And Suggestions.test.js which tests the overall ranking method. 

Some formatting changes to Categories to fix captialization errors. 

SuggestionTestData.js contains the test data for the suggestions class. 

Suggestions.js: fixes: query must take in a radius and a timeRange, getScore should check for undefined values rather than null, countCat needs a regular for loop in order to use break; and placeObjectsComparator needed its logic flipped (.sort uses ascending order).

### DIFF
--- a/step-capstone/src/__tests__/CountCatTests.test.js
+++ b/step-capstone/src/__tests__/CountCatTests.test.js
@@ -5,7 +5,7 @@ import {countCat} from '../scripts/Suggestion'
  * countCat testing
 */
 //one item matches preferences
-test("Counts type with one preference match correctly", () => {
+test("Place with a type matching a single user preference", () => {
   const userCat = ["familyFriendly", "outdoors"]
   const place = { 
     place_id: 0, 
@@ -16,7 +16,7 @@ test("Counts type with one preference match correctly", () => {
   expect(countCat(place, userCat)).toEqual(1)
 })
 
-test("Counts type with multiple preference matches correctly", () => {
+test("Place with a type matching mutliple user preferences", () => {
   const userCat = ["familyFriendly", "outdoors"]
   const place = { 
     place_id: 1, 
--- a/step-capstone/src/__tests__/CountCatTests.test.js
+++ b/step-capstone/src/__tests__/CountCatTests.test.js
@@ -1,0 +1,67 @@
+import '@testing-library/jest-dom'
+import {countCat} from '../scripts/Suggestion'
+
+/*
+ * countCat testing
+*/
+//one item matches preferences
+test("Counts type with one preference match correctly", () => {
+  const userCat = ["familyFriendly", "outdoors"]
+  const place = { 
+    place_id: 0, 
+    place: {
+      types: ["tourist_attraction", "aquarium"] 
+    }
+  } 
+  expect(countCat(place, userCat)).toEqual(1)
+})
+
+test("Counts type with multiple preference matches correctly", () => {
+  const userCat = ["familyFriendly", "outdoors"]
+  const place = { 
+    place_id: 1, 
+    place: {
+      types: ["tourist_attraction", "zoo"] 
+    }
+  } 
+  expect(countCat(place, userCat)).toEqual(2);
+})
+
+test("Zero matches with user preferences", () => {
+  const userCat = ["nightlife"]
+  const place = { 
+    place_id: 0, 
+    place: {
+      types: ["tourist_attraction", "aquarium"] 
+    }
+  } 
+  expect(countCat(place, userCat)).toEqual(0);
+})
+
+test("Doesn’t double count category matches", () => {
+  const userCat = ["sightseeing"]
+  const place = { 
+    place_id: 0, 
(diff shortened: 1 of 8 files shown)


step-capstone/src/__tests__/GetScoreTests.test.js
```diff
@@ -0,0 +1,74 @@
...
+import {getScore} from '../scripts/Suggestion'
+
+
+/*
+ * Score testing
+ */
+
+const ZERO_PROMINENCE = { total: 1, index: 1 }
+
+test("Handles category score correctly", () => {
```
ganewman42, line 11: Nit: These tests are not super descriptive, what is correct in this context?

step-capstone/src/__tests__/GetScoreTests.test.js
```diff
@@ -0,0 +1,74 @@
...
+
+test("Handles null Budget correctly", () => {
+  let score = getScore(1, 1, ZERO_PROMINENCE, undefined, 3, 0);
+  expect(score - 65).toEqual(5);
+})
+
+// test("Too expensive", () => {
+//   let score = getScore()
+//   expect(score - 65).toEqual(0);
+// })
```
ganewman42, line 69: Please remove or TODO

step-capstone/src/__tests__/SuggestionsTests.test.js
```diff
@@ -0,0 +1,144 @@
...
+  testPlaces.set(suggestionTestData.highRatedMountain.place.place_id, suggestionTestData.highRatedMountain);
+
+  const rankedResults = rank(testPlaces, config);
+
+  for (let i = 0; i < rankedResults.length - 1; i++) {
+    expect(rankedResults[i].score >= rankedResults[i + 1].score).toBeTruthy();
+  }
+})
+
+expect.extend({
```
ganewman42, line 130: Smart move to write your own equality checker.

step-capstone/src/__tests__/CountCatTests.test.js
```diff
@@ -0,0 +1,67 @@
+import '@testing-library/jest-dom'
+import {countCat} from '../scripts/Suggestion'
```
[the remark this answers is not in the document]
cthamrun: +1 on this. It would be ideal if we can make the mappings deterministic to avoid a source of flakiness as a best practice thing but I agree that it would be a nice-to-have instead of need-to-have for the purpose of the capstone.

step-capstone/src/__tests__/CountCatTests.test.js
```diff
@@ -0,0 +1,67 @@
...
+test("Counts type with multiple preference matches correctly", () => {
+  const userCat = ["familyFriendly", "outdoors"]
+  const place = { 
+    place_id: 1, 
+    place: {
+      types: ["tourist_attraction", "zoo"] 
+    }
+  } 
+  expect(countCat(place, userCat)).toEqual(2);
+})
```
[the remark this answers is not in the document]
zirving: Changed the doc string to make it clearer - this is testing that a place that has a type (in this case, zoo) that matches multiple user preferences (familyFriendly, outdoors), has both of those 'successes' counted. As opposed to aquarium, which matches familyFriendly, but not outdoors.

step-capstone/src/__tests__/CountCatTests.test.js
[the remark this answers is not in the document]
zirving: I agree. Added to our (growing) list of nice to have TODOS.

step-capstone/src/__tests__/GetScoreTests.test.js
zirving: This is checking the general use case of getScore, where some categories match. changed description for clarity

step-capstone/src/__tests__/GetScoreTests.test.js
zirving: implemented!

step-capstone/src/__tests__/GetScoreTests.test.js
hieudan225: I suggest "check getScore() function on a usual case".

step-capstone/src/__tests__/CountCatTests.test.js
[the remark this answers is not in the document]
hieudan225: I suggests the name for the first test: "One place_type matches one user category" and for the second test: "One place_type matches multiple user categories"

step-capstone/src/__tests__/CountCatTests.test.js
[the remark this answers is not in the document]
hieudan225: Good points. I think we can implement these ideas on week 11.

PR #35, opened 2020-07-17
Handle click in Timeline for suggestion request

hieudan225:
Adding a new feature: allow users to click in an empty slot in `Timeline` and a preference form will popup with information about the current empty time slot, the previous and subsequent `travelObject` items (if available).

Some notes for this PR:
1) I converted `TimeLine.js` to be a class component instead of a functional component, so that it can handle more complex features (in this PR is the `handleOnClickInterval` function, which needs to be binded with `this` of Timeline to work)
2) Important file:  `HandleClickedTimePoint.js` : 
a. function `getEmptySlots()` will always be called once the new props.list (a list of `TravelObjects`) has just been passed in `TimeLine`. This function costs O(the number of `travelObjects` are passed in `TimeLine`)
b. function `handleClickedTimePoint()` is a helper function for `handleOnClickInterval()` of `TimeLine.js`, which will handle the search for the current empty slot that the user click in, the previous and subsequent `travelObjects` of that empty slot
c. function `binarySearch()` is a helper function for `handleClickedTimePoint()`: does all the search and costs O( log(the number of `travelObjects` are passed in `TimeLine` )
3) Added file: `PreferencePopover.js`: will be modified to be a form for users to input their preferences (by Emmie). Currently, this form will popup when the users click in an empty slot. The information about the empty time slot (its time range, the previous and subsequent `travelObjects`) are being passed in as props to this form and ready to be used.
4) I made some fixes in `Suggestion.js` since I realized the Date parsing function has error.

Todo when merge code with Emmie: 
1) when users click submit the preference forms: produce a new Config object and pass that in the `getSuggestions` function. Discuss with Emmie and Zach whether `getSuggestions()` should be handled by `Trip.js` or by the `PreferencePopber` itself.
2) Finish the calculation of coordinates after Emmie hands over the coordinates.


step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
```diff
@@ -0,0 +1,75 @@
+/* Handling click into empty slot starts HERE */
+// Receives displayDate, returns the date string of displayDate
+const getDisplayDateString = (displayDate) => {
+  let year = displayDate.getFullYear().toString();
+  let month = (displayDate.getMonth() + 1 <10)? "0" + (displayDate.getMonth() + 1).toString() : displayDate.getMonth();
+  let date = (displayDate.getDate() < 10)? "0" + displayDate.getDate()   : displayDate.getDate();;
+  return year + "-" + month + "-" + date +   "T";
+};
```
ganewman42, line 8: I feel like there should be a way to use standardized `Date` functionality to do this. Looks like the `toString` method might be what you want.

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Receives displayItems list, returns a list of empty intervals
+export const getEmptySlots = (displayItems, displayDate) => {
+  let emptySlots = [];
+  const displayDateString = getDisplayDateString(displayDate);
+  if (displayItems.length == 0) {
+    emptySlots.push({ 
+        startDate: new Date(displayDateString + "00:00:00"),
```
ganewman42, line 16: In general you want to handle changes at the higher level data structure, rather than its string representation. This is because working with string representations can be pretty error prone (for example, maybe you miss a colon and your reviewers don't notice). So I would say you want to use date objects directly here and in several places below. You could do something like this:

```
startDate: new Date(displayDate.getFullYear(), displayDate.getMonth(), displayDate.getDate(), 0, 0, 0, 0)
endDate: new Date(displayDate.getFullYear(), displayDate.getMonth(), displayDate.getDate(), 23, 59, 59, 0)
```

step-capstone/src/components/Sidebars/PreferencePopover.js
```diff
@@ -0,0 +1,66 @@
...
+      } else if (props.slots.prevSlot !== null) {
+        coordinates.lat = props.slots.prevSlot.lat;
+        coordinates.lgn = props.slots.prevSlot.lgn;
+      } else if (props.slots.nextSlot !== null ) {
+        coordinates.lat = props.slots.nextSlot.lat;
+        coordinates.lgn = props.slots.nextSlot.lgn;
+      } else {
+        //TODO: implement when merge code with Emmie's preference form and autocomplete
+        //coordinates = props.trip_coordinates;
+      }
```
ganewman42, line 29: Nit: I know you haven't actually done this yet, but I think this else might not be quite right, since this should be the usual behavior, not the default behavior of using the surrounding coordinates.

step-capstone/src/scripts/Suggestion.js
```diff
@@ -1,9 +1,12 @@
...
 
-const getDateString = (dateObject)  => {
-    return dateObject.getFullYear() + "-" + dateObject.getMonth() + "-" + dateObject.getDate();
-}
+const getDisplayDateString = (displayDate) => {
+    let year = displayDate.getFullYear().toString();
+    let month = (displayDate.getMonth() <10)? "0" + displayDate.getMonth() : displayDate.getMonth();
+    let date = (displayDate.getDate() < 10)? "0" + displayDate.getDate()   : displayDate.getDate();;
+    return year + "-" + month + "-" + date;
+  };
```
ganewman42, line 9: As above, I think you should be able to use a library method for this.

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
```diff
@@ -0,0 +1,75 @@
...
+};
+
+/*Return the empty timeRange that users click into by performing binary search on emptySlots array*/
+export const handleClickedTimePoint = (idV, displayDate, emptySlots, displayItems) => {
+  let clickedTime = new Date(getDisplayDateString(displayDate) + idV + ":00");
+  let curSlot = binarySearch(emptySlots, clickedTime);
+  let prevSlot = binarySearch(displayItems, curSlot.startDate);
+  let nextSlot = binarySearch(displayItems, curSlot.endDate);
+  
+  return {curSlot: curSlot, prevSlot: prevSlot, nextSlot: nextSlot};
```
emaohn, line 74: What is prevSlot, and nextSlot? If binary search returns the empty time interval surrounding the clicked location, isn't that all we need?

step-capstone/src/components/Sidebars/PreferencePopover.js
```diff
@@ -0,0 +1,66 @@
...
+        coordinates.lat = (props.slots.prevSlot.lat + props.slots.nextSlot.lat)/2;
+        coordinates.lgn = (props.slots.prevSlot.lgn + props.slots.nextSlot.lgn)/2
+      } else if (props.slots.prevSlot !== null) {
+        coordinates.lat = props.slots.prevSlot.lat;
+        coordinates.lgn = props.slots.prevSlot.lgn;
+      } else if (props.slots.nextSlot !== null ) {
+        coordinates.lat = props.slots.nextSlot.lat;
+        coordinates.lgn = props.slots.nextSlot.lgn;
+      } else {
+        //TODO: implement when merge code with Emmie's preference form and autocomplete
```
emaohn, line 27: My component actually is the suggestion bar popup at the bottom 😅

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
emaohn: Oh I see they're the objects themselves. Maybe rename them for clarity? For example, timerange, prevTravelObject, nextTravelObject

step-capstone/src/components/Sidebars/PreferencePopover.js
emaohn: I created a function in Trip that can be passed down to the timeline to pass up the selected trip information into the suggestions bar since it also needs userPreferences.

emaohn: Could you perhaps make the getting coordinates a function in the timeline component? that way when I pass the function down to you all you need to pass up is the coordinates, radius, and timerange

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
hieudan225: good idea!

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
```diff
@@ -0,0 +1,75 @@
...
+    } else {
+      low = centerIndex + 1;
+    }
+  }
+
+  return null;
+};
+
+/*Return the empty timeRange that users click into by performing binary search on emptySlots array*/
+export const handleClickedTimePoint = (idV, displayDate, emptySlots, displayItems) => {
```
zirving, line 68: I am probably misunderstanding something, but could all this be simplified? If we have displayItems, all we need to do is find where displayItems[i].startDate>clickedTime? That would be the endDate of our empty interval and the prior item's endDate would be our startDate. Why do we need to search through a seperate list of empty slots?

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
hieudan225: Good point. I will change it accordingly to your suggestion.

step-capstone/src/components/Sidebars/PreferencePopover.js
hieudan225: I am not so clear about this comment. Could you kindly rephrase it for me?

step-capstone/src/components/Sidebars/PreferencePopover.js
hieudan225: Sure!

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
hieudan225: You're right, Zach. I can implement it the way you said, I'll just need to handle the case of where the user clicks in a completely empty timeline or the last time slot of the day by adding a fake travelobject at 23:59. When I implemented it, I was just thinking that those base cases were much, but now thinking back, it's not much at all. Okay, I will change this algorithm accordingly to your suggestion. 👍

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
```diff
@@ -0,0 +1,176 @@
...
+      startDate: displayItems[displayItems.length - 1].endDate,
+      endDate: endOfDisplayDate,
+    });
+  }
+
+  return emptySlots;
+};
+
+const contains = (interval, timePoint) => {
+  // return whether timeRange from startDate to endDate contains timePoint
```
cthamrun, line 37: Nit: move this comment to be above the function for consistency within the file.

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
```diff
@@ -0,0 +1,176 @@
...
+
+  return null;
+};
+
+/*Receives a list of travelobjects and returns the travelobjects that is in front and after the timePoint
+  fakeStart    |----A------|  timePoint  |--B--| |-C--|  |-------D------|         fakeEnd
+  return {A, {startDate: B}
+  if timePoint is in a travelobjects, return undefined for both A and B
+  if either A or B doesn't exist, return null for that travelobject
+*/
```
cthamrun, line 72: nit: have this as normal "//" comments for consistency within the file.

step-capstone/src/components/Sidebars/PreferencePopover.js
ganewman42: Yeah sorry that wasn't super clear :D What I was trying to say is that maybe the TODO should occur in a different case. However, I realized that I was wrong anyway so all good.

PR #36, opened 2020-07-17
Enable users to select preferences

emaohn:
- Autocomplete now returns placeId along with address and coordinates
- travelObjects now store placeId
- Trip stores a default coordinate for main destination
- Form component for adding user preferences
- Add user preference form to trip setting form
- Everything saves and retrieves properly with firebase
- Fixed minor bugs with map and indentation errors

Screen Shot 2020-07-17 at 12 27 06 PM

### DIFF
--- a/step-capstone/src/components/Pages/TripList.js
+++ b/step-capstone/src/components/Pages/TripList.js
@@ -49,7 +49,6 @@ class TripList extends React.Component{
   }
 
   handleAddTrip(newTrip) {
-    console.log(newTrip)
     this.context.addTrip(this.state.reference, newTrip);
     this.loadTrips();
   }
--- a/step-capstone/src/components/Pages/TripList.js
+++ b/step-capstone/src/components/Pages/TripList.js
@@ -49,6 +49,7 @@ class TripList extends React.Component{
   }
 
   handleAddTrip(newTrip) {
+    console.log(newTrip)
     this.context.addTrip(this.state.reference, newTrip);
     this.loadTrips();
   }
--- a/step-capstone/src/components/TravelObjectForms/AddEvent.js
+++ b/step-capstone/src/components/TravelObjectForms/AddEvent.js
@@ -21,7 +21,7 @@ export default function AddEvent(props) {
     const [endDate, setEndDate] = useState(overwriting ? props.data.endDate : props.startDate);
     const [checked, setChecked] = useState(overwriting ? props.data.finalized : false);
     const [title, setTitle] = useState(overwriting ? props.data.title : "");
-    const [location, setLocation] = useState(overwriting ? { address: props.data.location, coordinates: props.data.coordinates } : { address: null, coordinates: null });
+    const [location, setLocation] = useState(overwriting ? { address: props.data.location, coordinates: props.data.coordinates, placeId: props.data.placeId } : null);
     const [description, setDescription] = useState(overwriting ? props.data.description : "");
 
     const handleChecked = (e) => {
@@ -61,9 +61,10 @@ export default function AddEvent(props) {
             startDate: startDate,
             endDate: endDate,
             finalized: checked,
-            location: location.address,
-            coordinates: location.coordinates,
-            description: description
+            location: location ? location.address : null,
+            coordinates: location ? location.coordinates : null,
+            description: description,
+            placeId: location ? location.placeId : null
         })
         //validating input
         if (title === "") {
@@ -120,8 +121,8 @@ export default function AddEvent(props) {
                 <Box my={1}>
                     <LocationAutocompleteInput
                         onLocationSelected={handleLocationChange}
-                        error={(checked && location.address === null)}
-                        text={location.address}
+                        error={(checked && !location)}
+                        text={location ? location.address : ""}
                         type="event"
                     />
                 </Box>
--- a/step-capstone/src/components/TravelObjectForms/AddHotel.js
+++ b/step-capstone/src/components/TravelObjectForms/AddHotel.js
@@ -22,7 +22,7 @@ export default function AddHotel(props) {
     const [endDate, setEndDate] = useState(overwriting ? props.data.endDate : props.startDate);
     const [checked, setChecked] = useState(overwriting ? props.data.finalized : false);
     const [title, setTitle] = useState(overwriting ? props.data.title : "");
-    const [location, setLocation] = useState(overwriting ? { address: props.data.location, coordinates: props.data.coordinates } : { address: null, coordinates: null });
+    const [location, setLocation] = useState(overwriting ? { address: props.data.location, coordinates: props.data.coordinates, placeId: props.data.placeId} : null);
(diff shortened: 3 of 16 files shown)


ganewman42:
Please PR against where you branched off of (looks like `SuggestionTests` or `QueryRankingFixes`) rather than master to make the diff easier to read.

step-capstone/src/components/Pages/TripList.js
```diff
@@ -49,6 +49,7 @@ class TripList extends React.Component{
   }
 
   handleAddTrip(newTrip) {
+    console.log(newTrip)
```
hieudan225, line 52: you may forget to delete this console.log

step-capstone/src/components/TravelObjectForms/AddEvent.js
```diff
@@ -61,9 +61,10 @@ export default function AddEvent(props) {
             startDate: startDate,
             endDate: endDate,
             finalized: checked,
-            location: location.address,
-            coordinates: location.coordinates,
-            description: description
+            location: location ? location.address : null,
```
hieudan225, line 64: maybe rename this  variable `location` to `address` to distinguish it with  the location object which contains address?
so this line of code will be `address: location? location.address: null`

step-capstone/src/components/TravelObjects/TravelObject.js
```diff
@@ -12,7 +12,10 @@ export default function TravelObject(props) {
     
     const handleClick = (event) => {
         setAnchorEl(event.currentTarget);
-        props.onClickObject(props.data.id);
+        // if has a placeId, zoom to marker on map
```
hieudan225, line 15: great! so this will fix the problem we had last time: when user clicks in an object without an address, map breaks?

step-capstone/src/components/TripForms/AddTrip.js
```diff
@@ -24,28 +24,33 @@ const useStyles = makeStyles({
...
     const newTrip = {
         title: "",
         startDate: new Date(),
         endDate: new Date(),
-        destinations: "",
+        destination: null,
         description: "",
+        userPref: {
+            budget: 2,
+            radius: 10,
```
hieudan225, line 37: default radius is 10 meters?

emaohn: It's in kilometers for the sake of the slider haha. I put the slider in kilometers because a max of 50000 meters is a lot of increments, i just convert the km into m when getting suggestions

step-capstone/src/components/TravelObjects/TravelObject.js
emaohn: Yes! :D

step-capstone/src/components/TravelObjectForms/AddEvent.js
emaohn: Good point, will do!

step-capstone/src/components/TripForms/AddTrip.js
hieudan225: great!

step-capstone/src/components/TravelObjectForms/AddHotel.js
```diff
@@ -22,7 +22,7 @@ export default function AddHotel(props) {
     const [endDate, setEndDate] = useState(overwriting ? props.data.endDate : props.startDate);
     const [checked, setChecked] = useState(overwriting ? props.data.finalized : false);
     const [title, setTitle] = useState(overwriting ? props.data.title : "");
-    const [location, setLocation] = useState(overwriting ? { address: props.data.location, coordinates: props.data.coordinates } : { address: null, coordinates: null });
+    const [location, setLocation] = useState(overwriting ? { address: props.data.location, coordinates: props.data.coordinates, placeId: props.data.placeId} : null);
```
zirving, line 25: nit: can placeId not be part of location object - it doesnt seem to really fit with the purely location components of the place like address and coordinates. here and in addEvent I believe.

step-capstone/src/components/TripForms/TripSettingFormPopover.js
```diff
@@ -128,17 +140,15 @@ function EditTripSetting(props) {
...
-
                 <Grid item container direction="row" justify="space-between"></Grid>
                 <LocationAutocompleteInput
-                    onLocationSelected={handleDestinationsChange}
-                    error={destinations === ""}
-                    helperText={(destinations === "") ? "Cannot leave field blank, list your destinations, and separate them using comma" : null}
-                    text={destinations}
+                    onLocationSelected={handleDestinationChange}
+                    error={!destination}
+                    helperText={!destination ? "Cannot leave field blank, list your destinations, and separate them using comma" : null}
```
zirving, line 147: we should change this helperText since we are not supporting multiple destinations

step-capstone/src/components/Utilities/PreferenceForm.js
```diff
@@ -0,0 +1,260 @@
...
+                                        onChange={this.handleBudgetChange}
+                                        displayEmpty
+                                    >
+                                        <MenuItem value={0}></MenuItem>
+                                        <MenuItem value={1}>$</MenuItem>
+                                        <MenuItem value={2}>$$</MenuItem>
+                                        <MenuItem value={3}>$$$</MenuItem>
+                                        <MenuItem value={4}>$$$$</MenuItem>
+                                    </Select>
+                                    <FormHelperText>We won't give you anything above this price!</FormHelperText>
```
zirving, line 83: technically we will give them things above the price right? just later in the suggestions? maybe we can change the message? otherwise, the budget is probably fairly self explanatory.

step-capstone/src/components/TravelObjectForms/AddHotel.js
emaohn: Right now, location is storing the result of the autocomplete, which is an object containing all three--do you want for me to return {placeID, {address, coordinates} instead? Also, they're stored separately in the database (I realized I forgot to do it in AddHotel, but it's there in useEffect in AddEvent and AddFlight), I just keep them together in the form state for easy manipulation when the autocomplete location is changed/selected.

step-capstone/src/components/TripForms/TripSettingFormPopover.js
emaohn: Nice catch

step-capstone/src/components/Utilities/PreferenceForm.js
emaohn: True! Probably don't need helper text

step-capstone/src/components/TravelObjectForms/AddHotel.js
zirving: i think its fine as is since its all coming from the autocomplete and is stored seperately in the database. i just didnt want to have to do travelObject.location.placeID, but if its just in the form, its ok.

PR #37, opened 2020-07-17
Added Suggestion bar

zirving:
- Switched ItemPopover from using props.data==undefined to determine whether this was an edit or an add to using props.isNewItem: also changed FormPopover, TravelObject, and AddItemButton to pass in the new prop. 

- SuggestionBar.js: The overall bar that will display users suggestions. Contains two arrow buttons for next and previous which change a startindex for the displayed SuggestionItems, which get their data from the inputted suggestions array using that startindex.

- SuggestionItem.js: each suggestion item is a materialUI Card component which contains a header (the name of the place), an image from the place.photos (supplied by Google for most places, displays our logo otherwise). Also displays the Price using the $$ system and the Rating from 0 to 5. Finally, contains an Add button that launches the FormPopover and preloads it with the data from the suggestion. 

image

### DIFF
--- a/step-capstone/src/components/Suggestions/SuggestionBar.js
+++ b/step-capstone/src/components/Suggestions/SuggestionBar.js
@@ -10,7 +10,8 @@ export default class SuggestionBar extends React.Component {
 
     this.state = {
       suggestions: props.suggestions,
-      startIndex: 0
+      startIndex: 0,
+      numOfSuggestionsToDisplay: 3
     }
 
     this.loadPrevious = this.loadPrevious.bind(this);
@@ -30,12 +31,25 @@ export default class SuggestionBar extends React.Component {
 
 
   render() {
+    let suggestionItems = [];
+    for (let i = 0; i < this.state.numOfSuggestionsToDisplay; i++) {
+      suggestionItems.push(
+        <Grid item className="suggestion-item-holder">
+          <SuggestionItem
+            suggestion={this.state.suggestions[this.state.startIndex + i]}
+            context={this.props.context}
+            onAddItem={this.props.onAddItem}
+          />
+        </Grid>
+      );
+    }
+    
     return (
       <div id="suggestion-bar">
         <Grid id="suggestion-grid" container direction="row" wrap="nowrap" justify="center" >
           <Grid item>
             <IconButton
-              className="arrow-buttons" 
+              className="arrow-buttons"
               onClick={this.loadPrevious}
               disabled={(this.state.startIndex === 0) ? true : false}
             >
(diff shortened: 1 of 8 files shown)


step-capstone/src/components/Suggestions/SuggestionBar.js
```diff
@@ -0,0 +1,83 @@
...
+              />
+            </Grid>
+            <Grid item className="suggestion-item-holder">
+              <SuggestionItem 
+                suggestion={this.state.suggestions[this.state.startIndex + 2]} 
+                context={this.props.context}
+                onAddItem={this.props.onAddItem}
+              />
+            </Grid>
+          </Grid>
```
ganewman42, line 68: This seems a little awkward. What if you wanted to display 2 items instead, or 5? You'd have to copy-paste these blocks each time you make a change. If you can, a generator loop would be cleaner here.

step-capstone/src/components/Suggestions/SuggestionItem.js
```diff
@@ -0,0 +1,110 @@
...
+import logo from "../../images/logo.png"
+
+export default function SuggestionItem(props) {
+
+  const [anchorEl, setAnchorEl] = React.useState(null);
+
+  const handleClick = (event) => {
+    setAnchorEl(event.currentTarget);
+
+
```
ganewman42, line 23: Nit: Excess whitespace.

step-capstone/src/components/Suggestions/SuggestionItem.js
```diff
@@ -0,0 +1,110 @@
...
+        title="suggestion photo"
+      />
+
+      <CardContent>
+        <Grid container justify="space-between">  
+          <Typography className="suggestion-body" variant="subtitle2" inline gutterBottom align="left">
+            Price: {printDollarSigns(props.suggestion.place.price_level)} 
+          </Typography>
+          <Typography className="suggestion-body" variant="subtitle2" inline gutterBottom align="right">
+            Rating: {(props.suggestion.place.rating === undefined) ? "" : props.suggestion.place.rating}
```
ganewman42, line 80: Stretch goal: Displaying the rating as a number of stars or dots out of five might be a bit more intuitive. Currently, I get a number value, but I don't know how high the max is so it doesn't necessarily mean much to me.

hieudan225: or we can display the rating like this: 4.3 / 5 ?

emaohn: I think Dan's suggestion is pretty good for now. Places API might have some ui component that can give us a nice star-rating look (I know yelp does but not sure about Google)

step-capstone/src/components/Suggestions/SuggestionBar.js
emaohn: We could have a page index, which, multiplied by the number of places we want to display at a time would give us the position of the first result on that page and loop from that index however many places we want to display at a time. That way in the case we do want to change it all we have to change is the number we want to display.

emaohn: sorry that was super wordy haha

zirving: good idea all, it is now added, thanks!

step-capstone/src/components/Suggestions/SuggestionItem.js
zirving: fixed

step-capstone/src/components/Suggestions/SuggestionItem.js
zirving: good ideas. for now, I have just set it to display /5 like Dan suggested. Stars and dots get hard with decimals like .3 so I will leave that as a stretch goal.

PR #38, opened 2020-07-17
Suggestion popup

emaohn:
- Idea button that triggers popup for suggestions --> animated from bottom
- By default loads saved user preferences (from trip settings) into preference bar on side, but users can adjust and filter as they wish without changing their default preferences
- Display user preference bar on side for users to filter through suggestions on the spot (scrollable)
- Tab controller to display either activity suggestions or food suggestions --> will render Zach's component below the tab bar passing in whichever array is needed.
- Loading animation when querying places from API
- Retrieves suggestions based on user preferences (either passed in or specified) and ensures everything is well formatted to be passed into Zach's component
- Changed trip setting form from popover to modal in order to properly render in center of screen
- Trip functions are setup to be passed down to timeline to retrieve proper parameters for retrieving suggestions (radius, coordinates, time-range) and should be able to integrate with Dan's component through that.

Screen Shot 2020-07-17 at 5 20 49 PM
Screen Shot 2020-07-17 at 5 21 04 PM
Screen Shot 2020-07-17 at 5 23 15 PM

### DIFF
--- a/step-capstone/src/components/Pages/Trip.js
+++ b/step-capstone/src/components/Pages/Trip.js
@@ -175,6 +175,7 @@ export default class Trip extends React.Component {
         })
     }
 
+    // Triggered when a time slot is selected in timeline
     handleSelectTimeslot(timeRange, coordinates, radius) {
         this.setState({
             selectedTimeslot: {
@@ -185,6 +186,7 @@ export default class Trip extends React.Component {
         })
     }
 
+    // only allows rendering of suggestion bar once map is set
     renderSuggestionBar() {
         if (this.state.map) {
             return (
--- a/step-capstone/src/components/Pages/Trip.js
+++ b/step-capstone/src/components/Pages/Trip.js
@@ -2,12 +2,13 @@ import React from 'react';
 import Finalized from '../Sidebars/Finalized';
 import Unfinalized from '../Sidebars/Unfinalized';
 import AddItemButton from '../TravelObjectForms/AddItemButton'
-import { Grid, Box } from '@material-ui/core'
+import { Grid, Box} from '@material-ui/core'
 import '../../styles/Trip.css'
 import { FirebaseContext } from '../Firebase';
 import MapComponent from "../Utilities/Map"
 import { handleSuggestions } from "../../scripts/Suggestions"
 import GetSuggestionButton from '../Utilities/GetSuggestionButton';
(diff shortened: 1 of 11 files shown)


step-capstone/src/styles/SuggestionPopup.css
```diff
@@ -0,0 +1,20 @@
...
+    height: 350px;
+}
+
+.scroll {
+    overflow-y: auto;
+}
+
+.test {
+    background-color: aquamarine;
+}
```
zirving, line 16: I think this is left in accidentally?

emaohn: Yep, thanks for catching it!

step-capstone/src/components/Pages/Trip.js
```diff
@@ -33,7 +26,10 @@ export default class Trip extends React.Component {
             },
             map: null,
             service: null,
-            queryResults: null
+            queryResults: null,
+            palceIds: new Set(),
```
ganewman42, line 30: Misspelling, should be `placeIds`

step-capstone/src/components/Utilities/SuggestionPopup.js
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        this.handleUserPrefChange = this.handleUserPrefChange.bind(this);
+        this.handleToggleTab = this.handleToggleTab.bind(this);
+        this.renderSuggestions = this.renderSuggestions.bind(this);
+        this.getSuggestions = this.getSuggestions.bind(this);
+    }
+
+    componentDidMount() {
+        // this.getSuggestions();
```
ganewman42, line 40: Nit: Please remove or replace with a TODO. If this function should do nothing, there should be some sort of pass you can use.

step-capstone/src/components/Utilities/SuggestionPopup.js
```diff
@@ -0,0 +1,191 @@
...
+        if (!this.state.suggestionsLoaded) {
+            return (
+                <Grid container justify="center" align-items="center">
+                    <Box mt={10}>
+                        <CircularProgress color="secondary" />
+                    </Box>
+                </Grid>
+            )
+        }
+        // TODO: Suggestion component
```
ganewman42, line 120: Nit: What does this TODO mean? Isn't this the suggestion component?

PR #40, opened 2020-07-20
Static Testing

zirving:
snapshot testing works by rendering the react component and then checking that against a saved snapshot file. that is why the long snapshots need to be in the PR. 

it checks only that the created file is the same as the stored snapshot, so, if the code is purposefully made to render differently, you just update the snapshot. 

- created snapshot tests for the travelObjects, trip, tripList, and Home. 

- created mock for sessionStorage get and set methods: runs in setupTests.js which happens before any tests get run. these dont do anything for now, just allow code to run with test Data. in future they could be used to switch between test trips. 

- instead of Firebase API, created firebaseMock() which contains get methods  used by trip and tripList. you can also add any firebase functions you want to mock to this file.


step-capstone/src/__tests__/TripTests.test.js
```diff
@@ -0,0 +1,20 @@
...
+  sessionStorage.setItem("userId", "testUser");
+  sessionStorage.setItem("tripId", "testTrip");
+  
+  const tripComponent = renderer.create(
+    <FirebaseContext.Provider value={new firebaseMock()}>
+      <Trip/>
+    </FirebaseContext.Provider>
+    
+  );
+  await new Promise(res => setTimeout(() => { res() }, 0.5))
```
ganewman42, line 17: Nit: Please add a similar comment anyplace you had to put a wait, since that's kind of unusual for tests.

zirving: done

zirving:
thanks Grace, added directions to README

PR #42, opened 2020-07-22
Suggestion merge fixes

hieudan225:
In this PR, we fix these bugs:
1. `OneHourInterval` and its subcomponents (`TravelObject`, `ItemForm`) have their own independent onclick functions. When user clicks in `TravelObject` or `ItemForm` of a `OneHourInterval` div, the click event bubbles up to the parent component. Thus, the `InsertObjectPopover` always show up when we click in a `TravelObject`. We fix this problem by adding an if statement in `OneHourInterval's` onclick function: if the event is triggered by 'Interval', then call `props.onClickInterval` function; else do nothing.

2. Place auto completion doesn't send signal to its parent component when the users make any changes to the address without selecting a valid address. We fix this by calling the parent's `onLocationSelected(null)` everytime the location is changed.

3. When user clicks in a div twice, the `InsertObjectPopover` doesn't show up the second time. This problem is because `OneHourInterval` isn't notified when `InsertObjectPopover` closes, when `InsertObjectPopover` takes the anchorEl from `OneHourInterval`. We fix this problem by lifting the `state` and `setState` function for `anchorEl` of `InsertObjectPopover` to its parent component (`OneHourInterval`).

4. Fetching `photoUrl` issue: the service isn't get passed in to `TripSettingFormPopover` of `Trip`. We fix this problem by creating a map and service in `TripSettingFormPopover` instead.

5. Handling edge cases of location inside `Add[TravelObject]`: `location` can be `null`, an `empty string`, or an object with `address`, `placeId`, and `coordinates` being `null`.

6. Dynamically generate the appropriate `numOfSuggestionsToDisplay` based on the window size.

7. Coordinates of flight should not be considered on `getCoordinates` for suggestions.

8. UI fixes and other minor bugs


step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
```diff
@@ -159,7 +157,8 @@ export const handleClickedTimePoint = (idV, startOfDisplayDate, endOfDisplayDate
 
   let timePointEnd = new Date(timePointStart);
   timePointEnd.setTime(timePointEnd.getTime() + 30 * 60000);
-  
+  console.log(timePointStart)
+  console.log(timePointEnd)
```
ganewman42, line 161: Nit: Either add context to these or remove if they were meant to be temporary.

step-capstone/src/components/Sidebars/InsertObjectPopover.js
```diff
@@ -0,0 +1,95 @@
...
+  useEffect(() => {
+    if (props.anchorEl !== anchorEl) {
+      setAnchorEl(props.anchorEl);
+    }
+  }, [props.anchorEl]);
+
+  const getCoordinates = () => {
+    let coordinates = { lat: 0, lng: 0 }
+    if (props.slots !== null) {
+      if (props.slots.prevTravelObject !== null && props.slots.nextTravelObject !== null && props.slots.prevTravelObject.type !== 'flight' && props.slots.nextTravelObject.type !== 'flight') {
```
ganewman42, line 18: Nit: For super long criteria, it can be more readable if you pull out vars named after the  conditions as below. You can then re-use these for multiple criteria.

```
let prevNotNullOrFlight = props.slots.prevTravelObject !== null && props.slots.prevTravelObject.type !== 'flight'
let nextNotNullOrFlight = props.slots.nextTravelObject !== null && props.slots.nextTravelObject.type !== 'flight'
```

step-capstone/src/components/Sidebars/OneHourInterval.js
```diff
@@ -1,21 +1,31 @@
...
   const displayHeightOfADiv = 45.0;
   const padding = 22.0;
   const minPerDiv = 30.0;
+  const [anchorEl, setAnchorEl] = React.useState(null);
 
   const handleOnClickInterval = (idV, event) => {
-    let data = props.onClickInterval(idV);
-    if (data.freeTimeSlot !== undefined) {
-      setSlots(data);
+    console.log(event.target);
```
ganewman42, line 14: Nit: Similarly, add context or remove.

step-capstone/src/components/TravelObjectForms/ItemForm.js
```diff
@@ -51,6 +51,18 @@ export default class ItemForm extends React.Component {
     });
   }
 
+  /* Mimic Google Calendar's behavior: when user edits the newStartDate to be bigger than the current endDate,
+    this function will automatically reset the endDate to be bigger than the newStartDate by the same duration as before editting.
+    This behavior happends during editting, so the user can never submit an invalid time range.
```
ganewman42, line 56: Super nit: Spelling, happends->happens, editting->editing

step-capstone/src/components/Sidebars/HandleClickedTimePoint.js
zirving: removed, thanks

step-capstone/src/components/Sidebars/InsertObjectPopover.js
zirving: done, great idea, thanks

step-capstone/src/components/Sidebars/OneHourInterval.js
zirving: removed, thanks

step-capstone/src/components/TravelObjectForms/ItemForm.js
zirving: fixed, thanks

PR #43, opened 2020-07-23
Optimize routes branches

emaohn:
- getOptimalRoute function: given list of travel objects, returns them in an optimal order for travel
- createSchedule function: give ordered list of travel objects, generates a schedule taking into account the set durations of the travel objects, start and end times of the day (set in user preferences), and eating times (set in user preferences)
- editMultipleItems function: edits and saves a list of travel objects


step-capstone/src/scripts/Optimization.js
```diff
@@ -0,0 +1,85 @@
...
+            return { location: new window.google.maps.LatLng(travelObject.coordinates.lat, travelObject.coordinates.lng), stopover: true }
+        }),
+        travelMode: "DRIVING",
+        optimizeWaypoints: true
+    }
+    var directionsService = new window.google.maps.DirectionsService();
+
+    return new Promise(res => {
+        directionsService.route(request, function (result, status) {
+            if (status === "OK") {
```
cthamrun, line 19: What if the status here is not "OK"? It would be good to add in some error-handling here or have a multiple retry loop to handle flaky random errors from the directions service.

step-capstone/src/scripts/Optimization.js
```diff
@@ -0,0 +1,85 @@
...
+                    travelObjectsInOrder[i].toNextLocation = result.routes[0].legs[i];
+                }
+
+                res(travelObjectsInOrder);
+            }
+        })
+    })
+}
+
+export const createSchedule = function (travelObjects, userPref, displayDate) {
```
ganewman42, line 36: This function is complex enough and isolated enough that it could probably benefit from a unit test. To do this you might need to further separate into a utility, but it should be worth it.

step-capstone/src/scripts/Optimization.js
zirving: added error handling

step-capstone/src/scripts/Optimization.js
zirving: added testing!

PR #44, opened 2020-07-24
Food time preferences

emaohn:
- FoodTimeForm: new child component of preference form with three sliders to select preferred duration for breakfast, lunch, and dinner to aid with our optimization algorithm. Sliders are in minutes.
- Updated trip settings to include foodTimeRanges attribute, which is an array of length 3 each containing the number of milliseconds of the respective duration (0: breakfast, 1: lunch, 2: dinner).
- Default everything to 1 hour
- Resized setting popover form height and added scroll.

Screen Shot 2020-07-24 at 11 45 30 AM

### DIFF
--- a/step-capstone/src/components/Utilities/FoodTimeForm.js
+++ b/step-capstone/src/components/Utilities/FoodTimeForm.js
@@ -1,0 +1,107 @@
+import React, { useEffect, useState } from "react"
+import {
+    Grid,
+    Box,
+    TextField,
+    InputAdornment,
+    Input,
+    FormControl,
+    FormHelperText,
+    Slider,
+    Typography
+} from "@material-ui/core"
+const MILLISECONDS_PER_MIN = 60000;
+
+export default function FoodTimeForm(props) {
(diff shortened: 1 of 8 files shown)


step-capstone/src/components/Utilities/FoodTimeForm.js
```diff
@@ -0,0 +1,105 @@
+import React, { useEffect, useState } from "react"
+import {
+    Grid,
+    Box,
+    Slider,
+    Typography
+} from "@material-ui/core"
+
+export default function FoodTimeForm(props) {
+    const MILLISECONDS_PER_MIN = 60000;
```
ganewman42, line 10: Shouldn't there be a constant in the standard time library for this?

emaohn: I was unable to find one, but there are libraries like [Moment.js](https://momentjs.com/docs/#/durations/milliseconds/) that has conversion functions. Should I switch to using those?

PR #45, opened 2020-07-24
confirmation display for optimized schedule

zirving:
this confirmation displays the optimized schedule as a quick text list to let users see what we generated before they confirm they want it added to the schedule. 

image

### DIFF
--- a/step-capstone/src/components/Optimization/OptimizationConfirmation.js
+++ b/step-capstone/src/components/Optimization/OptimizationConfirmation.js
@@ -1,0 +1,58 @@
+import React, { useEffect, useState } from 'react';
+import {
+  Button,
+  Card,
+  CardActions,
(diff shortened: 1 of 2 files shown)


step-capstone/src/components/Optimization/OptimizationConfirmation.js
```diff
@@ -0,0 +1,58 @@
...
+    props.onConfirm(props.travelObjects);
+  }
+
+  useEffect(() => { setOpen(true) }, []);
+
+  return (
+    open ?
+      <Card className='optimization-confirmation'>
+        <CardContent>
+          <Typography variant='h5'> Here's what we have for you ...</Typography>
```
ganewman42, line 33: Random thing: I really like this wording.

PR #47, opened 2020-07-24
Prevent date conflict

zirving:
- The add and edit item forms now check whether the inputted dates overlap with an existing finalized event or flight. Hotels are ignored. 

- AddEvent and AddFlight now call hasTimeConflict to check with the about to be inputted item has a conflict. It passes in the id (to avoid a time conflict with the same travelObject, its start and end times, and the other travelObjects. 

- HelperFunctions.js has two new functions: hasTimeConflict iterates through the provided travelObjects and, if the object is finalized, and not the same object as the one being checked, calls the overlaps() function which determines if the times overlap. it returns true if there is a conflict

- All the other files have one single change - passing down the travelObjects list to allow the forms to check if there is a conflict.

### DIFF
--- a/step-capstone/src/components/TravelObjectForms/AddEvent.js
+++ b/step-capstone/src/components/TravelObjectForms/AddEvent.js
@@ -13,9 +13,7 @@ import {
 } from "@material-ui/pickers";
 import DateFnsUtils from "@date-io/date-fns";
 import LocationAutocompleteInput from "../Utilities/LocationAutocompleteInput";
-import {isValid, hasTimeConflict} from '../../scripts/HelperFunctions';
-
-const travelObjects = [{startDate:new Date(2020,6,24,8), endDate: new Date(2020,6,24,12)}]
+import { isValid, hasTimeConflict } from '../../scripts/HelperFunctions';
 
 export default function AddEvent(props) {
   let overwriting = props.data !== undefined;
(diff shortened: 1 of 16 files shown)


step-capstone/src/components/TravelObjectForms/AddEvent.js
```diff
@@ -134,6 +134,8 @@ export default function AddEvent(props) {
             label={"Start"}
             value={startDate}
             onChange={setStartDate}
+            error={checked && hasTimeConflict(props.data.id, startDate, endDate, props.travelObjects)}
+            helperText={checked && hasTimeConflict(props.data.id, startDate, endDate, props.travelObjects) ? "You already have something scheduled for that time" : ""}
```
ganewman42, line 138: Nice to have: In error messages, it's nice to specify what they should do to fix it. Just adding "Please choose another time or move {name of event}" would be a nice touch.

step-capstone/src/scripts/HelperFunctions.js
```diff
@@ -21,20 +22,42 @@ export const isValid = function (date) {
...
     )
   })
 }
 
+//returns true if there is a conflict, false otherwise
+export const hasTimeConflict = (id, startDate, endDate, travelObjects) => {
+  var foundConflict = false;
+  travelObjects.forEach((travelObject) => {
+    console.log("editing id", id)
+    console.log("travelobj", travelObject.id)
```
ganewman42, line 47: Please make these more clear or remove them.

zirving: removed, thanks

step-capstone/src/components/TravelObjectForms/AddEvent.js
zirving: good idea, added to nice to haves

PR #49, opened 2020-07-27
Drag and drop

hieudan225:
For this feature, I attempted two libraries:
1) `react-draggable`:
pros: can drag and scroll at the same time
cons: doesn't have drop target; can't have conditional drop; item at 12:00 AM will be dragged to unwanted places( I attempted to constraint bounds` but it doesn't work at all)

2) `react-dnd`:
pros: can have conditional drop; item will only be dropped if it is hovered upon drop target; has opacity setting for preview
cons: can't drag and scroll at the same time; if users place cursor at the tail of the item, preview doesn't quite match the drop (previewOptions which I hoped to place preview item in the correct position relative to the cursor doesn't work)

The current code is using `react-dnd` which seem to do it job but it's disappointing that we need to add another complicated library `react-dnd-scrollzone` just to get the scrolling works with dragging. Since we don't have much time, I decide to leave this feature at this stage for now to work on the next task and anyone of us can come back and fix this when they have time.


ganewman42:
Even before I read the code, nice work on knowing when to stop working on something, that can be harder than knowing when to start.

step-capstone/src/scripts/DragTravelObject.js
```diff
@@ -0,0 +1,56 @@
...
+    item.displayDate.getMonth(),
+    item.displayDate.getDate(),
+    23,
+    59,
+    59
+  );
+  var newStartDate = new Date(
+    item.data.startDate.getTime() +
+      (((diffPos) * minPerDiv) / pixelPerDiv) *
+        60000
```
ganewman42, line 32: Nit: This calculation is a little confusing, what's the 60,000? Maybe add an explanatory comment or make it clearer by using additional named variables.

hieudan225: Thank you, I have addressed this confusion in the latest commit.

PR #50, opened 2020-07-27
Change default date

zirving:
- Trip.js: pass in the date the user is currently looking at into add item rather than the start of the trip 

- Timeline.js: added updating of the display start and end dates to fix bug where adding items after start date of trip always used the same start date.

### DIFF
--- a/step-capstone/src/components/Sidebars/TimeLine.js
+++ b/step-capstone/src/components/Sidebars/TimeLine.js
@@ -66,9 +66,26 @@ export default class TimeLine extends React.Component {
   getIntervals() {
     let intervals = [];
     if (this.props.displayDate !== undefined) {
+      this.startOfDisplayDate = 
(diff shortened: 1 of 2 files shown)


step-capstone/src/components/Sidebars/TimeLine.js
```diff
@@ -66,9 +66,26 @@ export default class TimeLine extends React.Component {
   getIntervals() {
     let intervals = [];
     if (this.props.displayDate !== undefined) {
+      this.startOfDisplayDate =
```
ganewman42, line 69: Dangling equals? Might be a bug

zirving: it is actually just formatting, it is on the second line for some strange reason. it gets fixed on my other PR (fixUnboundedSuggestions)

PR #51, opened 2020-07-27
Final map features

emaohn:
- Clears and rerender's route when appropriate (deleting, adding, editing)
- Set default behaviors for map orientation:
Nothing in map: Set center to default destination coordinates
Changes in travel objects in current display date: set bounds to that day's route
Changes in other locations (finalized, unfinalized): set bounds to all markers
- Don't include departure airport in bounds
- Other bug fixes with clicking and responsive behavior
- Clicking marker pops up mini info window with title of travel object
- Added comparator for lists of travel objects and individual travel objects

### DIFF
--- a/step-capstone/src/components/TravelObjects/TravelObject.js
+++ b/step-capstone/src/components/TravelObjects/TravelObject.js
@@ -12,10 +12,8 @@ export default function TravelObject(props) {
     
     const handleClick = (event) => {
         setAnchorEl(event.currentTarget);
-        // if has a placeId, zoom to marker on map
-        if (props.data.placeId) {
-            props.onClickObject(props.data.id);
-        }
+
+        props.onClickObject(props.data.id);
     };
 
     const handleClose = (event) => {
--- a/step-capstone/src/components/Utilities/Map.js
+++ b/step-capstone/src/components/Utilities/Map.js
@@ -1,9 +1,8 @@
 import React, { createRef } from 'react'
 import { sortTravelObjectsByDate, sameTravelObjectList } from "../../scripts/HelperFunctions"
-import ZoomOutMapIcon from '@material-ui/icons/ZoomOutMap';
-import MapIcon from '@material-ui/icons/Map';
-import TodayIcon from '@material-ui/icons/Today';
-import { Typography } from '@material-ui/core';
+import ZoomInIcon from '@material-ui/icons/ZoomIn';
+import ZoomOutIcon from '@material-ui/icons/ZoomOut';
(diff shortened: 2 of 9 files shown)


step-capstone/src/components/TravelObjects/TravelObject.js
```diff
@@ -12,10 +12,8 @@ export default function TravelObject(props) {
     
     const handleClick = (event) => {
         setAnchorEl(event.currentTarget);
-        // if has a placeId, zoom to marker on map
```
ganewman42, line 15: Curious: What made this check no longer necessary?

step-capstone/src/components/Utilities/Map.js
```diff
@@ -40,24 +37,41 @@ class MapComponent extends React.Component {
...
+
+      this.zoomToMarker(marker);
+    } else if (!sameTravelObjectList(prevProps.displayDate.events, this.props.displayDate.events) && this.props.displayDate.events.length !== 0) {
+      // if changes to current day's events and isn't empty, zoom to todays events
       this.clearMap()
       this.drawMap();
       this.googleMap.fitBounds(this.getTodaysBounds());
-    } else if (prevProps.finalized !== this.props.finalized || prevProps.unfinalized !== this.props.unfinalized) {
+    } else if (!sameTravelObjectList(prevProps.finalized, this.props.finalized) || !sameTravelObjectList(prevProps.unfinalized, this.props.unfinalized)) {
+      // any other changes to finalized or unfinalized zoom out to encompass all markers.
```
ganewman42, line 48: Nice work on the comments.

step-capstone/src/scripts/HelperFunctions.js
```diff
@@ -21,20 +21,61 @@ export const isValid = function (date) {
...
+    }
+  }
+  return true;
+}
+
+export const sortTravelObjectsByDate = function (list) {
+  return list.sort(travelObjectStartDateComparator)
+}
+
+const equalsTravelObject = function (obj1, obj2) {
```
ganewman42, line 61: Nice work, this a good concept of equality.

step-capstone/src/components/TravelObjects/TravelObject.js
emaohn: I have a check in the map component that only zooms to the marker if it exists in the map (key is placeId)

emaohn: Actually never mind, I did need it--thanks for catching!

PR #63, opened 2020-07-27
Merge fix for user testing

emaohn:
- fixed some minor bugs from the merge
- exclude hotels from conflict detection
- adjust zoom to be a little more zoomed out
- display date retrieval bug
- remove whitespace from auto capitalization function return
- add error message to hotel date fields if same date
- fix double clicking when deleting trips

### DIFF
--- a/step-capstone/src/components/Pages/Trip.js
+++ b/step-capstone/src/components/Pages/Trip.js
@@ -289,7 +289,6 @@ export default class Trip extends React.Component {
         if (!this.state.items || !this.state.tripSetting) {
             return null;
         }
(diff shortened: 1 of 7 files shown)


step-capstone/src/components/Pages/Trip.js
```diff
@@ -289,6 +289,7 @@ export default class Trip extends React.Component {
         if (!this.state.items || !this.state.tripSetting) {
             return null;
         }
+        console.log(this.state.today.date)
```
ganewman42, line 292: Nit: Please remove old logging statement

PR #65, opened 2020-07-28
Error display for Optimization Form

hieudan225:
image
This RQ adds `ErrorDisplay` component for Optimization Form.


step-capstone/src/scripts/DropArea.js
```diff
@@ -65,6 +77,9 @@ export default function DropArea(props) {
     //     Call Dan's error display component here
     // }
     // })
+
+    
+    // onOpenError("Please remove some items!")
```
ganewman42, line 82: Two things:
1. Why is this commented out? Is that correct?
2. You don't say WHY the user has to remove items - are there too many things in the schedule?

hieudan225: Hi Grace, thanks for your comment. 1) It is commented out because it is just an example of how to use `ErrorDisplay` component.  2) The error message is actually generated in the `optimization.js` file. This PR is only about creating the `ErrorDisplay` component which takes in and displays the error message from the function using it (which is the function from `optimization.js` in this case).

PR #66, opened 2020-07-28
Replace share button with edit

zirving:
- Fixed a bug in TripSettingFormPopover where title was not being checked in useEffect(), so it didnt trigger a save. Also, we were not checking for valid dates in validation. 

- Added the Edit button to each TripItemComponent which calls handleEditTripClick() in TripList. This updates the selectedTripSettings and Id, which handleEditTrip uses on user save of the popover to make the edit call to firebase. Also added the Dialog/Popover combo we use for editing TripSettings, activated by handleEditTripClick()

### DIFF
--- a/step-capstone/src/components/Pages/TripList.js
+++ b/step-capstone/src/components/Pages/TripList.js
@@ -84,7 +84,7 @@ class TripList extends React.Component {
 
   handleEditTripSetting(newSetting) {
     this.context.editTripSetting(this.state.reference + this.state.selectedTripId, this.state.selectedTripSetting, newSetting)
-      .then(() => { console.log("load trips"); this.loadTrips() })
+      .then(() => { this.loadTrips() })
   }
 
   handleCloseTripSetting() {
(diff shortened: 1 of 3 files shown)


step-capstone/src/components/Pages/TripList.js
```diff
@@ -57,43 +64,85 @@ class TripList extends React.Component {
...
+          selectedTripSetting: tripSetting,
+          selectedTripId: tripId,
+          editingTrip: true
+        });
+      })
+  }
+
+  handleEditTripSetting(newSetting) {
+    this.context.editTripSetting(this.state.reference + this.state.selectedTripId, this.state.selectedTripSetting, newSetting)
+      .then(() => { console.log("load trips"); this.loadTrips() })
```
ganewman42, line 87: Nit: Please make the log message more specific or remove.

zirving: removed, thanks

PR #67, opened 2020-07-28
Get distance between coordinates for suggestions

emaohn:
- getDistance: function that takes two coordinates and a travel mode, returns distance and duration of travel
- When clicking between two travel objects, we get a radius based on the distance between the two points so that we get suggestions within the two.


step-capstone/src/components/Pages/Trip.js
```diff
@@ -261,17 +262,18 @@ export default class Trip extends React.Component {
         todayEndTime.setHours(23, 59, 59);
 
         if (this.state.map) {
+            console.log(this.state.selectedTimeslot && this.state.selectedTimeslot.radius ? this.state.selectedTimeslot.radius : this.state.tripSetting.userPref.radius)
```
hieudan225: you may want to remove this

step-capstone/src/components/Sidebars/TimeLine.js
```diff
@@ -37,6 +37,10 @@ export default class TimeLine extends React.Component {
   componentDidUpdate(prevProps) {
     if (prevProps.displayDate.toDateString() !== this.props.displayDate.toDateString) {
```
hieudan225, line 38: should it be:
if (prevProps.displayDate.toDateString() !== this.props.displayDate.toDateString())?

step-capstone/src/components/Sidebars/InsertObjectPopover.js
```diff
@@ -12,29 +13,46 @@ export default function InsertObjectPopover(props) {
...
-    }
+    })
+  }
 
-    return coordinates;
+  const calculateRadius = (start, end) => {
+    return new Promise(res => {
+      getDistance(start, end, "DRIVING").then((result) => {
+        let distance = result.distance.value;
+        // radius is half distance between two points + additional 2 km for encompassing start and endpoints.
```
hieudan225, line 51: you may want to modify the comments since 2km is no longer relevant, isn't it?

zirving, line 51: still have 2 km in the comment here even though gone from code

PR #72, opened 2020-07-29
Add hotel to start and end of route when applicable

emaohn:
- If stayed at hotel overnight, starts path with hotel
- If staying at hotel that night, ends path with hotel
- Ignore checkin/checkout hotel times in route


step-capstone/src/components/Utilities/Map.js
```diff
@@ -177,27 +177,34 @@ class MapComponent extends React.Component {
 
   // splits finalized trips into segments of coordinate pairs
   getPathSegments() {
+    let todaysHotel = this.props.date2HotelMap.get(this.props.displayDate.date)
+    console.log(todaysHotel)
```
zirving, line 181: nit: remove console.log

PR #78, opened 2020-07-31
Refactor HandleClickInTimeLine to handle edge cases & Add Unit Tests

hieudan225:
1) On writing test for `handle click in Timeline` feature, I realize the current code can't handle cases where one 30-min div has a small TravelObject in the middle of the div.
For example:
image

If user clicks in part A or part B of the div, the current code doesn't know which part is clicked and just return a random one A or B.
Before, we thought that using a function which looks for overlapping between the clicked div and a array of free Timeslot would help, but it only helps in cases like this (meaning overlapping function only helps deal with cases where a 30-min div has only one free timeslot):

image

Thus, this PR:
1) adds function to calculate the exact timePoint the user clicks into instead. (See `OneHourInterval`)
2) removes the `getFreeTimeSlot` and `searchForOverlappingPoint` function
3) adds unit tests to test the handleClickedTimePoint function. (See `HandleClickedTimePoint`)

(1) and (2) help reduce the time complexity of the `handle click in Timeline` feature from O(n) to O(logn) (where n is the number of displayItems), since we no longer need to iterate through displayItems to get the free timeslots anymore.

### DIFF
--- a/step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
+++ b/step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
@@ -43,6 +43,7 @@ const TIMEPOINT_1746 = new Date("2020-08-07T17:46:00Z");
 
 const TIMEPOINT_1800 = new Date("2020-08-07T18:00:00Z");
 const TIMEPOINT_2245 = new Date("2020-08-07T22:45:00Z");
+const TIMEPOINT_2246 = new Date("2020-08-07T22:46:00Z");
 
 const TIMEPOINT_2300 = new Date("2020-08-07T23:00:00Z");
 const TIMEPOINT_2330 = new Date("2020-08-07T23:30:00Z");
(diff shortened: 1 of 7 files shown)


step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
```diff
@@ -0,0 +1,666 @@
+import "@testing-library/jest-dom";
```
hieudan225, line 1: Please checkout this unit test file as well :D

step-capstone/src/components/Sidebars/OneHourInterval.js
```diff
@@ -21,9 +24,33 @@ export default function OneHourInterval(props) {
     }),
   });
 
+// Given a id of a div and an mouse click event
```
cthamrun: Nit: "Given an id of a div and a mouse click event"

step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
```diff
@@ -0,0 +1,666 @@
...
+      ONE_ITEM_IN_THE_MIDDLE
+    )
+  ).toStrictEqual({
+    prevTravelObject: undefined,
+    freeTimeSlot: undefined,
+    nextTravelObject: undefined,
+  });
+});
+
+test("Case 4: One travelobject overflows from the previous day and clickPoint is in the rest of the day", () => {
```
cthamrun: If I'm understanding this correctly, maybe this should be clarified as "clickPoint is in the part of the day after overflow event"

step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
```diff
@@ -0,0 +1,666 @@
...
+      ONE_ITEM_OVERFLOWS_FROM_PREV_DAY
+    )
+  ).toStrictEqual({
+    prevTravelObject: ITEM_2300_PREV_DAY_TO_0100,
+    freeTimeSlot: {startDate: TIMEPOINT_0100, endDate: endOfDisplayDate},
+    nextTravelObject: null,
+  });
+});
+
+test("Case 5: One travelobject overflows to the next day and clickPoint is in the rest of the day", () => {
```
cthamrun: If I'm understanding this correctly, maybe this should be clarified as "clickPoint is in the part of the day before overflow event"

step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
cthamrun: The test looks fine overall but there are some duplicate cases that looks like they could be eliminated. I see that there are a few foundational cases (clicking in between, clicking in, clicking before, clicking after an event, handling overflow events, etc) but it looks like the tests cover these for a different N number of events (2-5 events). If the cases are handled the same way regardless of N then it's probably better to have less tests for the same coverage.

zirving (approved):
LGTM

step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
hieudan225: Hi Chris, thank you for your comments! I was just thinking it'll be better to test the binary search algorithm with different array length as well as different search positions. (array in this case is displayItems) Can you tell me about your thoughts? @cthamrun

step-capstone/src/components/Sidebars/OneHourInterval.js
hieudan225: Thank you! Addressed :D

step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
hieudan225: Good suggestion! I'll rename the test like your suggestion.

step-capstone/src/__tests__/SearchForEmptyTimeslot.test.js
hieudan225: Good suggestion! I'll rename the test like your suggestion.

